PR #13, opened 2018-03-09
Update docs for Kubernetes 1.10

musse:


musse:
Questions so far:
- status of the `allow-privileged` flag
- do we want to only have info for K8s 1.10 or also for K8s 1.9?

saad-ali:
> what's the status of the allow-privileged flag?

kubeadm, gke, gce enabled, but default vanilla k8s it is disabled by default and must be explicitly enabled.

> do we want to only have info for K8s 1.10 or also for 1.9?

Move 1.9 docs in to an "archive". Complicated setup would be good not to lose it.

> Mount propagation beta with 1.10?

Correct.

saad-ali:
> List of drivers needs to be updated, what should we include

> Documentation on k8s.io link

Existing link should be fine

> Development section, what has changed between the 3 components?

Let's test it, and see what is different. 

> Deadline

March 26 -- 1.10 cut date.

musse:
Let's review and merge this and then I'll conclude with the testing section. @lpabon could you please take a look?

PR #19, opened 2018-03-28
Add GCE PD to drivers page

davidz627:
GCE PD is usable, I would like to add a link to the drivers page so it is more discoverable.


davidz627:
@saad-ali @lpabon PTAL

lpabon:
@davidz627 You have the output but not the markdown page added to the commit. You will loose the change on the next PR. Do you mind including the markdown page change?

davidz627:
fixed, thanks for the review

Issue #21, opened 2018-05-08
Add csi-digitalocean to the list of drivers


fatih:
Hi,

As per the instructions written at https://kubernetes-csi.github.io/docs/Drivers.html I would like to add our CSI plugin for DigitalOcean block storage to the list of production drivers.

The driver repo is here: https://github.com/digitalocean/csi-digitalocean
Name: `DigitalOcean Block Storage`
Status: `Alpha`
Description: `A Container Storage Interface (CSI) Driver for DigitalOcean Block Storage`

We're working on releasing the initial version `v0.0.1`. Until then it's marked as `alpha`. Please let me know if you need more information.

Thanks

PR #22, opened 2018-05-08
update drivers list

pohly:
This adds DigitalOcean Block Storage, as requested in
kubernetes-csi/docs#21. While at it, the list gets sorted
alphabetically because the current order seemed arbitrary.

### DIFF
--- a/book/src/Drivers.md
+++ b/book/src/Drivers.md
@@ -6,22 +6,22 @@ The following are a set of CSI driver which can be used with Kubernetes:
 ### Sample Drivers
 Name | Status | More Information
 -----|--------|-------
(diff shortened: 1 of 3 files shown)


fatih:
lgtm. Thanks @pohly

pohly:
Fatih Arslan <notifications@github.com> writes:

> Can we updated the version now to `v0.0.1` ?  We released a new
> version yesterday. It' still alpha but has a version now.

I've added that version to the PR as it is still open.

I don't know how closely the "drivers" page is meant to track version
numbers. Updating the page for every single driver release might get a
bit cumbersome over time.

fatih:
@pohly yeah makes sense. I've mentioned because it was still open. Also I agree, I think it's not worth writing the version going forward. It could give false permissions about the state of the driver if the driver is a new state but is not reflected here.

PR #29, opened 2018-06-11
Fix path for GCE PD CSI Driver

saad-ali:


saad-ali:
Oh, looks like I should be editing https://github.com/kubernetes-csi/docs/blob/master/book/src/Drivers.md not `docs/...`. Will close this PR and resubmit.

PR #33, opened 2018-07-25
Minor: Fix mount propagation link

sarjeet2013:
This PR fixes the (external) link for the mount propagation.


pohly:
Thanks for the fix. Before merging we need to also run `mdbook build` to update the .html.

@lpabon which version of mdbook are you using to generate the content? I tried with v0.1.3 but that alone led to some differences.

sarjeet2013:
@pohly Sure, I tried running `mdbook build` with `mdbook v0.1.8` and it changes lot of files. Not sure if this is intended or how this works? :)

Is there a specific version of `mdbook` this should be run with? Let me know.

lpabon:
@sarjeet2013 It's all good. As long as the .md files are ok

lpabon:
@sarjeet2013 @pohly Always use the latest mdbook. I don't really mind if there are lots of changes. I only review .md files

PR #45, opened 2018-08-29
fix misspelling

wnxn:
The author may want to write `kubelet plugin watcher socket` instead of `kubekel plugin watcher socket` .


lpabon:
Thanks! But do this needs to be fixed in the markdown sources: 
https://github.com/kubernetes-csi/docs/blob/master/book/src/Setup.md#kubelet-plugin-watcher 

Please also, build the new changes using `mdbook` as shown on the main README.md:
https://github.com/kubernetes-csi/docs#kubernetes-csi-documentation

lpabon:
No longer applicable, but thank you for the change.

PR #49, opened 2018-09-05
WIP Add documentation for snapshot

xing-yang:
This PR adds documentation for the snapshot feature.

Note that images for external-snapshotter, external-provisioner, and hostpath need to be replaced with official ones once they are uploaded.


xing-yang:
This PR will be replaced with another one that uses DaemonSet instead of a regular pod.

PR #61, opened 2018-10-08
fix misspelling

wnxn:
correct '--feature-dates' to '--feature-gates'


lpabon:
Thank you, this is fixed by #65

PR #65, opened 2018-10-15
Update README

leakingtapan:
For:
1. explicit state that feature gates need to be set for both kubelet and kube-api server. This took me two days to realize this is the issue. 
1. Update provisioner role permission for topology awareness scheduling.


book/src/Setup.md
```diff
@@ -2,10 +2,10 @@
 This document has been updated for the latest version of Kubernetes v1.12.  This document outlines the features that are available for CSI.  To get step by step instructions on how to run an example CSI driver, you can read the [Example](./Example.md) section.
 
 ## Enabling features
-Some of the features discussed here may be at different stages (alpha, beta, or GA).  Ensure that the feature you want to try is enabled for the Kubernetes release you are using.  To avoid version mismatch, you can enable all of the features discussed here with:
+Some of the features discussed here may be at different stages (alpha, beta, or GA).  Ensure that the feature you want to try is enabled for the Kubernetes release you are using.  To avoid version mismatch, you can enable all of the features discussed here for both kubelet and kube-apiserver with:
```
verult: Maybe replace it with "across all Kubernetes components"? Is it common to have different feature gates across components in the same cluster?

leakingtapan: "across all kubernetes components" is a vague and much broader term to me (it includes kube-scheduler, controller-manager, kube-proxy, etc). Especially for new people, they might not even know what are all the components. How about we keep it explicit as it is for now?

verult: OK. My original concern was we have to remember to update this comment if we end up depending on a flag in controller-manager, for example, but the likelihood of that is probably low.

PR #70, opened 2018-10-29
Added livenessprobe configuration example for CSI driver

sarjeet2013:
@msau42 @lpabon Please review.

I added the example configuration snippet from https://github.com/kubernetes-csi/livenessprobe#using-livenessprobe

### DIFF
--- a/book/src/Example.md
+++ b/book/src/Example.md
@@ -133,6 +133,66 @@ csi-hostpathplugin-9rp7c     2/2     Running   0          5s
 {{#include example/hostpath/csi-hostpathplugin.yaml}}
 ```
 
+
+### Deploy livenessprobe with CSI plugin
+
+The CSI community provides a livenessprobe side-container that can be integrated with the CSI driver services (Node and Controller) to provide the liveness of the CSI service containers.
+
+The livenessprobe side-container will expose the an http endpoint that will be used in a kubernetes liveness probe.
+
+Below is an example configuration which needs to be added to CSI driver services (Node and Controller) yamls:
+
+Note: This example is derived from [using-livenessprobe](https://github.com/kubernetes-csi/livenessprobe#using-livenessprobe) from [kubernetes-csi/livenessprobe](https://github.com/kubernetes-csi/livenessprobe)
+
+```yaml
+- name: hostpath-driver
+    image: quay.io/k8scsi/hostpathplugin:vx.x.x
+    imagePullPolicy: Always
+    securityContext:
+      privileged: true
+#
+# Defining port which will be used to GET plugin health status
+# 9808 is default, but can be changed.
+#
+    ports:
+    - containerPort: 9808
+      name: healthz
+      protocol: TCP
+    livenessProbe:
+      failureThreshold: 5
+      httpGet:
+        path: /healthz
+        port: healthz
+      initialDelaySeconds: 10
+      timeoutSeconds: 3
+      periodSeconds: 2
+      failureThreshold: 1
+...
+#
+# Spec for liveness probe sidecar container
+# 
+ - name: liveness-probe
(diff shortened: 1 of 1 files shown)


book/src/Example.md
```diff
@@ -133,6 +133,63 @@ csi-hostpathplugin-9rp7c     2/2     Running   0          5s
...
+      httpGet:
+        path: /healthz
+        port: healthz
+      initialDelaySeconds: 10
+      timeoutSeconds: 3
+      periodSeconds: 2
+      failureThreshold: 1
+...
+#
+# Spec for liveness probe sidecar container
```
msau42: I would add some more detail here about what is going on.  This sidecar container is exposing the /healthz endpoint that the livenessProbe above queries.

book/src/Example.md
```diff
@@ -133,6 +133,63 @@ csi-hostpathplugin-9rp7c     2/2     Running   0          5s
 {{#include example/hostpath/csi-hostpathplugin.yaml}}
 ```
 
+
+### Deploy livenessprobe with CSI plugin
+
+Below is an example configuration which needs to be added to CSI driver/provisioner yaml:
```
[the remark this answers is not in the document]
sarjeet2013: provisioner I meant was controller. 

Will reword it to: "which needs to be added to CSI driver and/or controller yaml." 

Will this be fine?

book/src/Example.md
sarjeet2013: Sure, will add.

sarjeet2013:
@msau42 Updated. PTAL

book/src/Example.md
```diff
@@ -136,7 +136,11 @@ csi-hostpathplugin-9rp7c     2/2     Running   0          5s
 
 ### Deploy livenessprobe with CSI plugin
 
-Below is an example configuration which needs to be added to CSI driver/provisioner yaml:
+The CSI community provides a livenessprobe side-container that can be integrated with the CSI driver services (Node and Controller) to provide the liveness of the CSI service containers.
+
+The livenessprobe side-container will enable the integration with kubernetes liveness Probe, and will expose the endpoint on a httpGet request for CSI driver services. The same port is configured as `health-port` for liveness-probe side-container.
```
msau42: Maybe reword to something like "The livenessprobe sidecar container will expose an http endpoint that be used in a Kubernetes liveness probe."

sarjeet2013: done

sarjeet2013:
@msau42 updated based on re-review comment.

book/src/Example.md
```diff
@@ -133,6 +133,66 @@ csi-hostpathplugin-9rp7c     2/2     Running   0          5s
...
+    args:
+    - --csi-address=/csi/csi.sock
+    - --connection-timeout=3s
+    - --health-port=9898
+#
+```
+
+Where:
+
+- `--csi-address` - specifies the Unix domain socket path, on the host, for the CSI driver. It allows the livenessprobe sidecar to communicate with the driver for driver liveness information.  Mount path `/csi` is mapped to HostPath entry `socket-dir` which is mapped to directory `/var/lib/kubelet/plugins/csi-hostpath`
```
msau42: so this is not actually the path on the host right? it's the path as seen in the container?

sarjeet2013: @msau42  Yes. It will be path which will be seen inside container by livenessprobe side-container.

msau42:
Thanks this lgtm, please squash all the commits

sarjeet2013:
@msau42  Thanks for reviewing. I've squashed all commits to one. Please take a look and merge if looks good?

saad-ali:
CC @sbezverk for FYI

/lgtm

PR #72, opened 2018-11-06
example: pull from upstream repos directly, use fixed revisions

pohly:
Maintaining local copies of all the necessary files is unnecessary, we
might as well tell kubectl to pull directly from the individual
repos. But to ensure that this really pulls the right revision of each
file, we have to specify the revision hashes instead of the "master"
branch.

Updating those hashes manually can be cumbersome. The new
hack/update-example.sh script automates that step. The remaining local
.yaml files are the ones that have no corresponding upstream file. It
might be worth pushing them upstream in the future, because then
everything will be in one place (hostpath deployment + examples)
and/or easier to update (first update in upstream repo, then update
docs with the hashes).


pohly:
/cc @msau42 
/hold

https://github.com/kubernetes-csi/external-snapshotter/pull/53 must be merged first, then the Example.md needs to be updated an the HTML files re-generated. But some early feedback would be welcome before I invest more time into this, and I wanted to have this PR open as explanation why that other PR is useful.

pohly:
/

https://github.com/kubernetes-csi/external-snapshotter/pull/53

hack/update-example.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# It's easier to update the .yaml files, submit a PR, and then
+# once accepted, update the example using this script.
+
+example=book/src/Example.md
+
+for url in $(grep 'kubectl create -f.*https://raw.githubusercontent.com/' $example | sed -e 's;.*\(https://raw.githubusercontent.com[^ (]*\).*;\1;'); do
+    # split URL into pieces
+    eval $(echo $url | sed -e 's/\]$//' -e 's;https://raw.githubusercontent.com/\([^/]*/[^/]*\)/\([^/]*\)/\(.*\);repo=\1 rev=\2 file=\3;')
+    git fetch https://github.com/$repo refs/heads/master:refs/remotes/$repo/master
```
msau42: Instead of latest, would it make more sense to pull specific tags from released versions?

j-griffith: @msau42 That makes more sense in a lot of ways, my concern though would be making sure we *remember* to update things like the docs repo.  That being said, it would be easy to just set a version file and update it with each CSI release and update that way.  Could even potentially automate  it at some point?

msau42: Yes, we could setup some fancy CI that automates a lot of this.  In the meantime, it will have to be one of those things to remember to do with every release.

pohly: I had considered a parameter to pull from a branch, but then didn't add that because I wasn't sure whether we would ever need it. If needed and easier than manually updating the Example.md, it can still be added at that time.

lpabon:
let's go with your change for now, and then add a page with "Releases" which has many versions of these pulled yamls

PR #73, opened 2018-11-08
Update urls in docs

cofyc:
`csi-api/pkg/crd/testdata` was renamed to `csi-api/pkg/crd/manifests` in https://github.com/kubernetes/kubernetes/pull/70193


pohly:
There is already a PR open for this: PR #72

cofyc:
oh, I missed it, using fixed revision is better.

Issue #78, opened 2018-11-13
Add Makefile for docs


saad-ali:
Updating docs is a two step process:

1. Update file in `book/*`
2. Run generation command to generate files in `docs/*`

We should add a Makefile at the root that makes it dead simple to run the command:

```
make all <-- generate all files
make clean <-- remove temp files
make clobber <-- remove all generated files
```

leakingtapan:
I was originally proposing adding makefile for my `docker run --rm -v $(pwd):/data -u $(id -u):$(id -g) -it chengpan/mdbook:0.2.1 mdbook build ./book
` command to generate the book. Seems @lpabon had some concerns on owning the makefile. So this did happen. Not sure if he changed his opinion?

PR #81, opened 2018-11-16
add Makefile

pohly:
"make all/clean/clobber/serve" can be used for building or viewing
documentation locally. The default is to download the mdbook binary
for Linux x86. This can be tuned with Makefile variables or a custom
binary can be used, see Makefile for detail.

Fixes: #78

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,50 @@
+ROOT := $(shell pwd)
+
+# The default is to download the official release binary
+# for x86_64-unknown-linux-gnu. If that default is unsuitable,
+# build your own binary and copy or symlink it here, or
+# override the variable with "make MDBOOK_BINARY=...".
+MDBOOK_BINARY = $(ROOT)/mdbook
+
+# The architecture can be changed. Only Linux .tar.gz files
+# are currently supported, though. See https://github.com/rust-lang-nursery/mdBook/releases
+# for available architectures.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,50 @@
...
+# The default is to download the official release binary
+# for x86_64-unknown-linux-gnu. If that default is unsuitable,
+# build your own binary and copy or symlink it here, or
+# override the variable with "make MDBOOK_BINARY=...".
+MDBOOK_BINARY = $(ROOT)/mdbook
+
+# The architecture can be changed. Only Linux .tar.gz files
+# are currently supported, though. See https://github.com/rust-lang-nursery/mdBook/releases
+# for available architectures.
+MDBOOK_ARCH = x86_64-unknown-linux-gnu
```
leakingtapan: How about using the docker image that I built `chengpan/mdbook:0.2.1` so that we don't need to worry about installing mdbook and cross platform issue? I will be happy to transfer it to some official repo if people has concerns it is a personal repo.

pohly: I did indeed have concerns about the maintenance of that docker image. After having looked at how difficult (or rather, easy) it is to just download the pre-build mdbook, that's the approach that I prefer personally.

pohly: Another reason is that for Travis CI, using Docker is just plain overkill.

PR #84, opened 2018-11-20
Hostpath driver 1.0 YAML update

vladimirvivien:
Updates documentation to point to YAML configs for 1.0.0 update.

### DIFF
--- a/book/src/Example.md
+++ b/book/src/Example.md
@@ -35,11 +35,13 @@ in v1.12.
 CRDs need to be created manually for `CSIDriverRegistry` and `CSINodeInfo`:
 
 
-> $ kubectl create -f [https://raw.githubusercontent.com/kubernetes/csi-api/ab0df28581235f5350f27ce9c27485850a3b2802/pkg/crd/testdata/csidriver.yaml](https://github.com/kubernetes/csi-api/blob/ab0df28581235f5350f27ce9c27485850a3b2802/pkg/crd/testdata/csidriver.yaml) --validate=false
-> `customresourcedefinition.apiextensions.k8s.io/csidrivers.csi.storage.k8s.io created`
+> $ kubectl create -f [https://github.com/kubernetes/csi-api/blob/dea832f47dd0116156b428802bc7079039c2456c/pkg/crd/manifests/csidriver.yaml](https://github.com/kubernetes/csi-api/blob/dea832f47dd0116156b428802bc7079039c2456c/pkg/crd/manifests/csidriver.yaml) --validate=false
+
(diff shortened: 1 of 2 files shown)


book/src/Example.md
```diff
@@ -91,7 +93,7 @@ The CSI sidecar apps are going to connect to the CSI driver, therefore
 starting it first helps avoid timeouts and intermittent container
 restarts:
 
-> $ kubectl create -f [https://raw.githubusercontent.com/kubernetes/kubernetes/f40a5d1155aae95105a4e9bb8933d750c666e350/test/e2e/testing-manifests/storage-csi/hostpath/hostpath/csi-hostpathplugin.yaml](https://github.com/kubernetes/kubernetes/blob/f40a5d1155aae95105a4e9bb8933d750c666e350/test/e2e/testing-manifests/storage-csi/hostpath/hostpath/csi-hostpathplugin.yaml)
+> $ kubectl create -f [https://github.com/kubernetes/kubernetes/blob/1a3fbf1f13a28f3102edc43ab2dca95471453cd8/test/e2e/testing-manifests/storage-csi/hostpath/hostpath/csi-hostpathplugin.yaml](https://github.com/kubernetes/kubernetes/blob/1a3fbf1f13a28f3102edc43ab2dca95471453cd8/test/e2e/testing-manifests/storage-csi/hostpath/hostpath/csi-hostpathplugin.yaml)
```
msau42: Instead of pulling from kubernetes e2e test manifests, should we pull from yamls in kubernetes-csi/drivers?

pohly:
The top of the README.md states that the example is known to work for 1.12. Before changing something on that page, please decide how people will find working examples for the different supported Kubernetes versions.

book/src/Example.md
pohly: That makes sense, but someone would first have to update the drivers repo such that it actually has up-to-date yaml files. Right now there aren't any.

msau42: Let's do that first, so we don't have to update this twice?

book/src/Setup.md
```diff
@@ -171,7 +171,7 @@ To configure your driver using the registrar sidecar, you can configure the cont
         fieldRef:
           apiVersion: v1
           fieldPath: spec.nodeName
-    image: quay.io/k8scsi/driver-registrar:v0.2.0
+    image: quay.io/k8scsi/driver-registrar:v1.0.0
```
saad-ali: Move this to 1.0.1
1.0.0 has a bug that will prevent it from working (it advertises the wrong supported version).

pohly: And replace driver-registrar with node-driver-registrar.

vladimirvivien:
Closing because this has been fixed elsewhere.

PR #86, opened 2018-11-23
Add GlusterFS to snapshot supported CSI Driver list.

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>


saad-ali:
/lgtm
/approve

We should modify the list in https://kubernetes-csi.github.io/docs/Drivers.html to be a table with capabilities (drivers and dynamic provisioning, volume snapshots, etc.).

Issue #104, opened 2019-01-28
Add CSI driver for XSKY


hoyho:
Hi there, we have finished our CSI driver now . 
vendor: XSKY
description: A Container Storage Interface (CSI) Driver for XSKY distributed block storage (X-EBS)
version: Beta
homepage: [https://www.xsky.com/en/](https://www.xsky.com/en/)
please help adding to table [Production Drivers](https://kubernetes-csi.github.io/docs/Drivers.html)
thanks

msau42:
please send a pr

PR #108, opened 2019-01-29
Add XSKY CSI driver to production drivers list

hoyho:
#104


hoyho:
Already signed a CLA

saad-ali:
Looks like commit is from `luohaihao` instead of `hoyho` that might be what's making the CLA bot upset.

hoyho:
yes , then I change the commit user to `hoyho`

hoyho:
I signed it

PR #111, opened 2019-02-01
Fix example hostPath mounts for csi-driver-registrar

mhulscher:
Fixes the YAML-definition of the hostPath-volumes used as an example in the csi-driver-registrar.

### DIFF
--- a/book/src/Setup.md
+++ b/book/src/Setup.md
@@ -181,14 +181,14 @@ To configure your driver using the registrar sidecar, you can configure the cont
       mountPath: /registration
 ...
 volumes:
+- hostPath:
+    path: /var/lib/kubelet/plugins/csi-hostpath
+    type: DirectoryOrCreate
   name: socket-dir
-  - hostPath:
-      path: /var/lib/kubelet/plugins/csi-hostpath
-      type: DirectoryOrCreate
+- hostPath:
+    path: /var/lib/kubelet/plugins_registry
+    type: Directory
   name: registration-dir
-  - hostPath:
-      path: /var/lib/kubelet/plugins_registry
-      type: Directory
 ```
 Where:
 


mhulscher:
signed the CLA

mhulscher:
Signed the CLA?

mhulscher:
Definitely signed it.

msau42:
Thanks for the PR! We just did a rewrite of the docs and I think what we want to do going forward is to:
* directly link to the documentation in the sidecar: this has been done as part of the rewrite
* link to a complete driver/yaml example from the examples.md page: this still needs to be done.  I think the hostpath csi driver would be a good example to link to.  cc @pohly for any inputs

pohly:
Basically all of the deployment commands from https://kubernetes-csi.github.io/docs/example.html need to be ripped out and replaced with documentation that shows how to use https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/deploy/deploy-hostpath.sh

Using the deployed driver then can use the YAML files from https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/examples

There is something for snapshotter (https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/deploy/snapshotter) but it looks like that part is currently incomplete (not deployed, no examples). That needs to be enhanced in csi-driver-host-path first.

pohly:
Let's not duplicate information, though. https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/README.md already explains how to create and use a PVC. Let's move the snapshot example completely into that same file and then in kubernetes-csi/docs just link to it with a few words, like this:

> The HostPath can be used to provision local storage in a single node test. The [driver's  documentation](https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/README.md) explains how to deploy and use that driver in Kubernetes.

> The deployment of a CSI driver determines which RBAC rules are needed. For example, enabling or disabling leadership election changes which permissions the external-attacher and external-provisioner need. The HostPath example deployment uses the original RBAC rule files that are maintained together with those sidecar apps and deploys into the default namespace.

> A real production should copy the RBAC files and customize them as explained in the comments of those files.

okartau:
> Let's move the snapshot example completely into that same file and then in kubernetes-csi/docs just link to it

I can try to make these changes, to get some touch and practice in this repo.

PR #122, opened 2019-03-02
Add Synology NAS driver to the list of drivers

jparklab:
I added to the list of sample drivers instead of production since I do not work for Synology and do not intend to support this driver for a production use.
I have been using this driver for my home lab.


jparklab:
I finished the CLA signature.

saad-ali:
Can you change the description from:

`A Container Storage Interface Driver for Synology NAS.` to `An unofficial (and unsupported) Container Storage Interface Driver for Synology NAS.`

jparklab:
Good point, I updated the description

Issue #124, opened 2019-03-06
Outdated and bad links for examples


vladimirvivien:
* [The example page](https://kubernetes-csi.github.io/docs/example.html) should be redone to point to new location of hostpath and hostpath doc https://github.com/kubernetes-csi/csi-driver-host-path/blob/master/README.md.

* The link to Hostpath driver here is not working https://kubernetes-csi.github.io/docs/deploying.html

pohly:
@okartau intents to work on that.

pohly:
My suggestion is that most of the text in the example page gets replaced with a short description that kubernetes-csi uses the hostpath driver as example, then link to the csi-driver-host-path README.md for details.

That README.md then has to cover deploying the driver with the deploy script and then can copy most of the usage examples from the current example page.

okartau:
liveness-probe is not covered by hostpath deployment, so this part remains documented here?

okartau:
> and then can copy most of the usage examples from the current example page.

with "and then can copy" part you mean that we repeat examples from this page in hostpath, or we move them, or we link to them?

msau42:
We should add liveness probe to hostpath so we can get e2e test coverage

okartau:
I created a 1st draft PR, simply deleted duplicate part.
Here is some snapshotter material which is not on hostpath, to be migrated there?
Also, liveness-probe to be added (and doc migrated) on hostpath side, as was already mentioned here.

pohly:
> with "and then can copy" part you mean that we repeat examples from this page in hostpath, or we move them, or we link to them?

I meant "move". There's no point in having duplicated content.

PR #125, opened 2019-03-07
example.md: reduce the contents which are present in hostpath deployment example

okartau:
This page was mostly repeating (older version) of what is written in hostpath
deployment example. We can point to hostpath and delete
the outdated deployment examples here.

Resolves: #124


book/src/example.md
```diff
@@ -3,136 +3,26 @@
 > ## *This page is out-of-date and under active development.*
 
 The
-[HostPath](https://github.com/kubernetes-csi/drivers/tree/master/pkg/hostpath)
-can be used to provision local storage in a single node test. This
-section shows how to deploy and use that driver in Kubernetes.
+[HostPath](https://github.com/kubernetes-csi/csi-driver-host-path)
```
msau42: "The Hostpath CSI driver is a simple sample driver that provisions a directory on the host. It can be used as an example to get started writing a driver, however it is not meant for production use."

book/src/example.md
```diff
@@ -3,136 +3,26 @@
...
+can be used to provision local storage. The deployment example there
```
msau42: "deployment example there": can you link to the actual page?

book/src/example.md
```diff
@@ -3,136 +3,26 @@
...
-
-> $ kubectl create -f [https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/01bd7f356e6718dee87914232d287631655bef1d/deploy/kubernetes/rbac.yaml](https://github.com/kubernetes-csi/external-snapshotter/blob/01bd7f356e6718dee87914232d287631655bef1d/deploy/kubernetes/rbac.yaml)
-> ```
-> serviceaccount/csi-snapshotter created
-> clusterrole.rbac.authorization.k8s.io/external-snapshotter-runner created
-> clusterrolebinding.rbac.authorization.k8s.io/csi-snapshotter-role created
-> ```
-
-### Deploy driver-registrar and hostpath CSI plugin in DaemonSet pod
+CRDs are created optionally `CSIDriverRegistry` and `CSINodeInfo`:
```
msau42: I don't know if we need this anymore considering we're getting rid of the CRDs in 1.14.

okartau: Let's remove CRD stmt from here, it does not add much value, glimpse into deployment example shows CRDs are inside if stmt, thus optional

book/src/example.md
okartau: I will make "deployment example there" to be a link to actual deployment script, if that's what you mean

book/src/example.md
okartau: yes, I will replace the sentence, and overall I think the first part can be simplified. The header "Deployment" should go away, and remaining sentences get reduced and for one chapter

okartau: it may be better idea to link to deploy/ directory instead of deployment script, as the whole set of files under deploy/ forms this example

okartau:
force-pushed latest state. Link to example now points to hostpath deploy/ dir.
What do you think, do we need to keep stmt "The CSI sidecar apps are going to connect..." ?

Next:
- liveness-probe deployment example to be moved from here to hostpath deployment script
  (needs deployment verification, I will try)
- remaining snapshot example should be moved there as well
- then those parts deleted from here

okartau:
liveness probe deployment added as PR#27 in hostpath repo

book/src/example.md
```diff
@@ -1,138 +1,13 @@
...
-> ```
->
-> $ kubectl get pod
-> ```
-> NAME                         READY   STATUS    RESTARTS   AGE
-> csi-hostpath-attacher-0      1/1     Running   0          14s
-> csi-hostpath-provisioner-0   1/1     Running   0          56s
-> csi-hostpathplugin-4k7hk     2/2     Running   0          117s
-> ```
+The CSI sidecar apps are going to connect to the CSI driver, therefore starting it first helps avoid timeouts and intermittent container restarts
```
msau42: I'm not sure what this sentence means and how it translates into something you can do in your Pod specs. Can you clarify?

book/src/example.md
```diff
@@ -192,177 +67,10 @@ Where:
...
->   Attacher:   csi-hostpath
->   Node Name:  127.0.0.1
->   Source:
->     Persistent Volume Name:  pvc-0571cc14-c714-11e8-8911-000c2967769a
-> Status:
->   Attached:  true
-> Events:      <none>
-> ```
-
 ## Snapshot support
```
msau42: @xing-yang do you want the snapshot example here, or should it go in the snapshot page and/or the snapshotter repo?

pohly: I would put it into the csi-driver-host-path repo.

* this page: can get out of sync with the actual example(s) - can be more than one, depending on the Kubernetes version
* snapshot page: same issue
* snapshotter repo: doesn't have a complete deployment, and I don't think we should add one there

book/src/example.md
pohly: Put the CSI driver container at the top of the list of containers in the pod, if that pod also contains the sidecards.

It's less of an issue now because we improved the connection behavior of the sidecars, but previously we've seen sidecars time out in their connection attempt because the CSI driver was started after them and image pulling took too long. Having said that, I still think it's a good advice.

book/src/example.md
pohly: We discussed this today in the Kubernetes-CSI meeting and agreed to use the csi-driver-host-path repo.

book/src/example.md
pohly: This is an advice about best-practices. Thinking about it some more, I now suggest that we remove the entire sentence. It doesn't belong into the example page.

okartau: I deleted that sentence and pushed new state

okartau:
The current PR state has addressed latest comments, right.
Next steps to get this moving further, a plan to complete this PR using isolated changes steps:
- get PR#27 in hostpath repo merged, adding liveness-probe in deploy example there
- after above merged, I will delete liveness-probe part in this PR here
- then we can deal with moving remaining snapshot example parts from here into hostpath example as well, in similar fashion. i.e. preparing PR there, merging it, then deleting respective parts in this PR here.

book/src/example.md
```diff
@@ -7,63 +7,7 @@ Deployment has been tested with Kubernetes v1.13
...
-```
-
-Where:
-
-- `--csi-address` - specifies the Unix domain socket path, as seen in the container, for the CSI driver. It allows the livenessprobe sidecar to communicate with the driver for driver liveness information.  Mount path `/csi` is mapped to HostPath entry `socket-dir` which is mapped to directory `/var/lib/kubelet/plugins/csi-hostpath`
-
-- `--connection-timeout` - specifies the timeout duration of waiting for CSI driver socket in seconds. (default 30s)
-
-- `--health-port` - specifies the TCP ports for listening healthz requests (default "9808")
+The example deployment includes livenessprobe side-container provided by the CSI community, using configuration derived from [using-livenessprobe](https://github.com/kubernetes-csi/livenessprobe#using-livenessprobe) from [kubernetes-csi/livenessprobe](https://github.com/kubernetes-csi/livenessprobe)
```
okartau: I re-used sentence from original, but the double "from" feels complicated here.
The second "from" just repeats the repository part of first "from", so what about we delete 2nd as it does not add new information.
Adding 2nd link would make sense if it pointed to some other repository.

okartau: supported by not from @pohly, I deleted 2nd "from" part and force pushed

okartau:
I created a https://github.com/kubernetes-csi/csi-driver-host-path/pull/31 that makes next step, migrates rest of snapshot example from here to hostpath deployment example

okartau:
https://github.com/kubernetes-csi/csi-driver-host-path/pull/31 was merged, so I pushed a new state here that replaces snapshot example contents with one sentence (please check it).

Note that there is now the last sentence "please check the Troubleshooting page" which has broken link. 
Do we want to have such link there (should fix it) or can we just delete this sentence now,
as we have moved away actual possibly trouble-creating parts from this page

pohly:
Please remove the sentence "Deployment has been tested with Kubernetes v1.13". This page no longer documents that.

> Note that there is now the last sentence "please check the Troubleshooting page" which has broken link.

The case must be fixed (`Troubleshooting.html` -> `troubleshooting.html`). The page itself isn't very useful anymore, but let's fix the link and then decide later whether that page still makes sense.

okartau:
1.13 sentence deleted and case in troubleshooting fixed (another fix is .md instead of .html), pushed.
(most of) troubleshooting file entries are getting old and/or are not relevant?

pohly:
> another fix is .md instead of .html

The `.html` suffix was correct, it's what we need in the actual site (https://kubernetes-csi.github.io/docs/example.html). Please revert to .html.

okartau:
oh, I naively hoped I can verify link here as well. Reverted-

book/src/example.md
```diff
@@ -1,493 +1,12 @@
 # Example
 
-> ## *This page is out-of-date and under active development.*
+The [Hostpath CSI driver](https://github.com/kubernetes-csi/csi-driver-host-path) is a simple sample driver that provisions a directory on the host. It can be used as an example to get started writing a driver, however it is not meant for production use.
+The [deployment example](https://github.com/kubernetes-csi/csi-driver-host-path/tree/master/deploy) shows how to deploy and use that driver in Kubernetes.
```
pohly: Sorry, only thought of this now: https://github.com/kubernetes-csi/csi-driver-host-path#deployment is a better link, because it has some actual documentation.

okartau: good idea! pointing to deployment chapter is more user-friendly than pointer to directory

book/src/example.md
```diff
@@ -1,493 +1,12 @@
...
-section shows how to deploy and use that driver in Kubernetes.
+The example deployment uses the original RBAC rule files that are maintained together with sidecar apps and deploys into the default namespace. A real production should copy the RBAC files and customize them as explained in the comments of those files.
 
-The deployment of a CSI driver determines which RBAC rules are
-needed. For example, enabling or disabling leadership election changes
-which permissions the external-attacher and external-provisioner need.
-This example deployment uses the original RBAC rule files that are
-maintained together with those sidecar apps and deploys into the
-default namespace.
+The example deployment includes livenessprobe side-container provided by the CSI community, using configuration derived from [using-livenessprobe](https://github.com/kubernetes-csi/livenessprobe#using-livenessprobe).
```
msau42: Do we need to explicitly call out liveness probe and snapshot?  We're not calling out any of the other sidecars.

okartau: current structure is result of mechanically replacing one chapter with one-two sentences about it. Other sidecars did not have dedicated chapters.

msau42: let's leave them out then.  The new docs page has separate pages for each of the sidecars that explain what they do.

okartau: we leave out livenessprobe and snapshot sentences, and we keep RBAC chapter?

msau42: yup I think that is fine

okartau: pushed it. The remaining text is not long :-)

msau42:
/lgtm
/approve
Thank you!

PR #121, opened 2019-03-01
Update CSIDriver and CSINodeInfo features to beta

jsafrane:
This should fix our docs for Kubernetes 1.14.

cc @xing-yang to check the API corresponds to the PR

/hold
Until https://github.com/kubernetes/kubernetes/pull/73883/ is merged

PR #128, opened 2019-03-13
fix text - missing D in CRD

shay-berman:


saad-ali:
/lgtm
/approve

@shay-berman could you make sure to sign the CNCF CLA. Once you've done that this should merge.

shay-berman:
I just changed the commit auther to be my ibm email (as mentioned [HERE](https://github.com/kubernetes/community/blob/master/CLA.md#4-ensure-your-github-e-mail-address-matches-address-used-to-sign-cla)).

Will keep you update when I will finish the CNCF CLA signing.

saad-ali:
@msau42 pointed out CR is correct here (Custom Resource not Custom Resource Definition). Maybe we can expand out the terms instead of using the acronym (and maybe link to the definitions) for clarity.

shay-berman:
1. So maybe it should be "object" instead of CR?  (the word 'object' mentioned in the begining of the sentence so it will be consistent)

2. In addition we can add CRD and link to the CSIDriver CRD in the beginning of the page 
"The CSIDriver Kubernetes API object [CRD](https://github.com/kubernetes/csi-api) serves two purposes"

Thoughts?

shay-berman:
I signed it

shay-berman:
@saad-ali waiting for your inputs.

saad-ali:
How about:

```
CSI drivers do not need to create the `CSIDriver` CRD or CR directly. Instead they may use the [cluster-driver-registrar](cluster-driver-registrar.md) sidecar container (customizing it as needed with startup parameters). When deployed with a CSI driver it automatically creates a `CSIDriver` CRD (if one doesn't already exist) and a `CSIDriver` CR to represent the driver.
```

This will have to be updated for 1.14.

msau42:
This entire section is going to change signficantly after https://github.com/kubernetes-csi/docs/pull/121

shay-berman:
@saad-ali is this PR still relevant or the https://github.com/kubernetes-csi/docs/pull/121 already covers it?

shay-berman:
Hi @saad-ali 
Not sure if this PR is relevant any more, since as @msau42  mentioned (and its already in the master branch for 1.14) the section was changed signficantly and fixed the CRD references.

So can i close this PR?

Suggestion - we may improve the text a bit (maybe as part of this PR) - and to add how the  `CSIDriver` object of the CSI plugin is removed from k8s. (Because it only mentioned how "What creates the CSIDriver object").
Do you think we should mentioned it to complete the story of how the CSIDriver object removed if the CSI plugin removed from the  cluster.

msau42:
@shay-berman I think that's a great idea

shay-berman:
ok @msau42

Can you please approve that the deletion process is as follow (and then i will update the PR accordingly):
"
What deletes the CSIDriver object of the plugin?
The CSIDriver object of the specific CSI plugin is automatically removed when the CSI plugin(csi-controller statefulset and csi-node daemonset) removed from the cluster. Since the cluster-driver-registrar sidecar is part of the CSI plugin, then k8s identify that there is no more CSI pluginX registered so it deletes the CSIDriver object of the CSI plugin.
"

msau42:
Nothing automatically removes the CSIDriver object.  Right now we're recommending that drivers manually include the spec as part of their deployment, so when uninstalling all the specs of the driver, the CSIDriver object will get deleted

PR #134, opened 2019-03-27
Document raw blocks

bswartz:
Explain how plugins should support raw blocks.

### DIFF
--- a/book/src/raw-block.md
+++ b/book/src/raw-block.md
@@ -1,8 +1,9 @@
-# Raw Block Volume Feature
+﻿# Raw Block Volume Feature
 
-> ## *This page is still under active development.*
+## Status
 
-**Status:** Alpha
+* Kubernetes 1.11 - 1.13: Alpha
+* Kubernetes 1.14: Beta
 
 This page documents how to implement raw block volume support to a CSI Driver.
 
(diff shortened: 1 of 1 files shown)


book/src/raw-block.md
```diff
@@ -2,7 +2,7 @@
 
 > ## *This page is still under active development.*
 
-**Status:** Alpha
+**Status:** Beta
```
msau42: Also note which K8s version it's beta in.  And that it also requires the nodes to be on 1.14 too

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
 
 The [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) supports both block and mounted (file) volumes.
 
-While Kubernetes support of mounted (file) volumes is GA/stable, support for [block volume in Kubernetes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support) was introduced in v1.9, and promoted to beta in Kubernetes 1.13.
+While Kubernetes support of mounted (file) volumes is GA/stable, support for [block volume in Kubernetes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#raw-block-volume-support) was introduced in v1.9, and promoted to beta in Kubernetes v1.13. The CSI-specific portions of block volumes was introduced in v1.11 promoted to beta in v1.14.
```
msau42: To avoid confusion, I think we should leave out the in-tree support versions and only talk about versions appropriate to CSI.

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
 
-The Kubernetes CSI tooling support for block volumes is still alpha.
+## Implementing Raw Block Volume Support
```
msau42: suffix with "in your CSI driver"

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+capabilities requested, or else return an error. If the `AccessType` method of a `VolumeCapability`
+`VolumeCapability_Block`, then the capability is requesting a raw block volume. Unlike mount volumes, block
+volumes don't have any specific capabilities that need to be validated, although access modes still
+apply.
+
+Block volumes are much more likely to support multi-node flavors of `VolumeCapability_AccessMode_Mode`
+than mount volumes, because there's no filesystem state stored on the node side that creates any technical
+impediments to multi-attaching block volumes. There may still be good reasons to prevent multi-attaching
+block volumes, and there may be implementations that are not capable of supporting multi-attach, but it's
+generally wrong to use exactly the same set of access modes for block as for mount volumes.
```
msau42: instead of generally wrong, I would probably just say something like "evaluate your driver appropriately"

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+
+CSI plugins that support both mount and block volumes must be sure to check the capabilities of all CSI RPC
+requests and ensure that the capability of the request matches the capability of the volume, to avoid trying
+to do file-system-related things to block volumes and block-related things to file system volumes.
+
+Also, CSI plugins that implement the optional `GetCapacity()` RPC should note that that RPC includes
+capabilities too, and if the capacity for mount volules is not the same as the capacity for block
```
msau42: typo: volumes

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+volumes don't have any specific capabilities that need to be validated, although access modes still
+apply.
+
+Block volumes are much more likely to support multi-node flavors of `VolumeCapability_AccessMode_Mode`
+than mount volumes, because there's no filesystem state stored on the node side that creates any technical
+impediments to multi-attaching block volumes. There may still be good reasons to prevent multi-attaching
+block volumes, and there may be implementations that are not capable of supporting multi-attach, but it's
+generally wrong to use exactly the same set of access modes for block as for mount volumes.
+
+CSI plugins that support both mount and block volumes must be sure to check the capabilities of all CSI RPC
```
msau42: would it be useful to list exactly which RPCs we pass in the capabilities to?

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+
+Q: Can CSI plugins support only block volumes and not mount volumes?
+A: Yes! This is just the reverse case of supporting mount volumes only. Plugins may return `InvalidArgument`
+for any creation request with an `AccessType` of `VolumeCapability_Mount`.
+
+## Differences Between Block and Mount Volumes
+
+The main difference between block volumes and mount volumes is the expected result of the `NodePublish()`.
+For mount volumes, the CO expects the result to be a mounted directory, at `TargetPath`. For block volumes,
+the CO expect there to be a device file at `TargetPath`. The device file can by a bind-mounted device from
```
msau42: typo: expects

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+the hosts `/dev` filesystem, or it can be a device node created at that location using `mknod()`.
```
msau42: This may also imply that your driver needs to mount a "/dev" hostpath

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+
+It's desirable but not required to expose an unfiltered device node. For example, CSI plugins based on
+technologoies that implement SCSI protocols should expect that pods consuming the block volumes they create
```
msau42: typo: technologies

book/src/raw-block.md
```diff
@@ -11,22 +11,62 @@ A *mounted (file) volume* is volume that will be mounted using a specified file
...
+only hard requirement is that the device implements block reading/writing however.
+
+For plugins with the `RPC_STAGE_UNSTAGE_VOLUME` capability, the CO doesn't care exactly what is placed at
+the `StagingTargetPath`, but it's worth noting that some CSI RPCs are allowed to pass the plugin either
+a staging path or a publish path, so it's important to think carefully about how `NodeStageVolume()` is
+implemented, knowing that either path could get used by the CO to refer to the volume later on. This is
+made more challenging because the CSI spec says that `StagingTargetPath` is always a directory even for
+block volumes.
 
-Because this feature is still alpha it is disabled by default in Kubernetes. You must enable the feature on Kubernetes:
```
msau42: We should still document the feature gates that control the behavior on Kubernetes even though it's enabled by default

book/src/raw-block.md
msau42: I don't know if we really need to mention it here since it's not really specific to block volumes, but the driver shouldn't assume that it's the only driver running on that node (and therefore owns all the devices).

bswartz:
Gah! I need a spell checker. Thanks for the review.

book/src/raw-block.md
bswartz: I'm not sure these are raw-block-specific issues. Nearly all CSI drivers of any type will need to have a hostpath mount of /dev or even /. Maybe this advice belongs in a more general section on tips for writing CSI plugins.

book/src/raw-block.md
```diff
@@ -1,8 +1,8 @@
-# Raw Block Volume Feature
+﻿# Raw Block Volume Feature
 
 > ## *This page is still under active development.*
 
-**Status:** Alpha
+**Status:** Beta since 1.14 (Alpha since 1.11)
```
msau42: K8s 1.14  (just so there's no confusion with sidecar versions)

msau42: Actually let's follow this format: https://github.com/kubernetes-csi/docs/pull/121/files#diff-64e1d803bf8d66339de7965984ce84ccR5

PR #145, opened 2019-04-13
Update sidecar image versions

msau42:


msau42:
/hold
wait until everything is released

book/src/external-attacher.md
```diff
@@ -6,8 +6,9 @@
 
 **Status:** GA/Stable
 
-Latests stable release | Branch | Compatible with CSI Version | Container Image | Min k8s Version | Max k8s version
+Latest stable release | Branch | Compatible with CSI Version | Container Image | Min k8s Version | Max k8s version
```
saad-ali: Discussed offline.

*Let's rename `Compatible with CSI Version` to `Min CSI Version`.
* And add a column for `Max CSI Version`

Max should be empty for everything except the v0.x.x releases.

msau42:
I also added a commit that rewords the feature pages to be more consistent with each other

book/src/external-attacher.md
```diff
@@ -6,10 +6,11 @@
 
 **Status:** GA/Stable
 
-Latests stable release | Branch | Compatible with CSI Version | Container Image | Min k8s Version | Max k8s version
+Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min k8s Version | Max k8s version
 --|--|--|--|--|--
```
xing-yang: Need another |-- ?

gnufied: Is Max CSI version or Max k8s version basically redundant fields? I do not see value in filling those out.

gnufied: In particular Max. CSI version means there was a breaking change and this sidecar is no longer compatible with that version of CSI. Is there a chance of this happening in post 1.0 CSI world?

msau42: There were two reasons I had in mind for a Max CSI version field:
1. CSI spec has a 2.0
2. CSI spec introduced some new features that we haven't implemented support for yet.  As a hypothetical example, let's say external-provisioner needed to add support for resizing in 1.1, but we didn't implement it it yet, but still released a sidecar version. The max would reflect that we're not using the capabilities of the new spec yet.

book/src/external-snapshotter.md
```diff
@@ -6,10 +6,11 @@
 
 **Status:** Alpha
 
-Latests stable release | Branch | Compatible with CSI Version | Container Image | Min k8s Version | Max k8s version
+Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min k8s Version | Max k8s version
 --|--|--|--|--|--
```
xing-yang: Missed a |--.  It is not displayed as a table because of that.

book/src/external-provisioner.md
```diff
@@ -6,10 +6,11 @@
 
 **Status:** GA/Stable
 
-Latests stable release | Branch | Compatible with CSI Version | Container Image | Min k8s Version | Max k8s version
+Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min k8s Version | Max k8s version
 --|--|--|--|--|--
```
xing-yang: Missed a |--.  It is not displayed as a table because of that.

msau42:
/hold
wait for new external-attacher cut

saad-ali:
/approve
/lgtm

Discussed offline. I'm a little worried that updating `Min k8s Version` may confuse people. @msau42 and I discussed maybe adding `Recommended k8s Version` column instead with a disclaimer under the table documenting the difference between minimum and recommended. And we can update the feature pages to have a table documenting not just k8s version but also sidecar container versioning per feature for alpha/beta/GA. Can do that in a follow up.

PR #159, opened 2019-05-20
Update the Ember-CSI driver

Akrog:
```release-note
NONE
```


saad-ali:
Do we really want to drop the `if storage backend supports it`?

Akrog:
> Do we really want to drop the `if storage backend supports it`?

Thank you for the careful consideration.

For Cinder drivers both, dynamic provisioning and snapshot support, are basic features that all drivers  must support, so it's not driver dependent.

saad-ali:
Ack.
/approve
/lgtm

PR #162, opened 2019-05-23
Add csi drvier name and expansion for YanRongYun

wpf710:
support expansion


saad-ali:
1. Blocker for this PR: can you please update the link to point to the CSI driver or instructions on how to install the CSI driver instead of a generic product page. 
2. Not a blocker for this PR, but FYI, the CSI driver name should be in domain name format.

PR #166, opened 2019-05-31
docs: update xsky driver

hoyho:
1. support xsky file storage
2. support snapshot for block storage


hoyho:
> 1. Can you please put yes or no for the last column, i.e. does driver support volume resize (expansion).
> 2. Not a blocker for this PR, but FYI, the CSI driver name should be domain notation.

updated

PR #168, opened 2019-06-01
Update secrets documentation with new keys and template variables

### LINKED ISSUE #150: Add more examples for secret templating
The docs currently have example StorageClasses for using only fixed secret references, or only templated secret references.

It might be useful to also add an example StorageClass that combines both fixed and templated references.  

And also describe some more sample user stories that this can support.

/help

ggriffiths:
Fixes #150 
- I believe there was a typo for Node Stage Secret and Controller Publish Secret docs. Only the name for a secret may be `${pvc.name}` or `${pvc.annotations}`, not the namespace:
https://github.com/kubernetes-csi/external-provisioner/blob/967b7a310b197d3294b52af84a195b45b9a0e1d2/pkg/controller/controller.go#L830-L839
- added a new example with secret provided for multiple operations.
- updated the provisioner secret template variables supported.

Release notes:
```release-notes
NONE
```

### DIFF
--- a/book/src/secrets-and-credentials.md
+++ b/book/src/secrets-and-credentials.md
@@ -48,10 +48,19 @@ The value of both parameters may be a literal or a template containing the follo
 
     * `${pv.name}`
       * Automatically replaced with the name of the `PersistentVolume` object being provisioned at provision.
+    * `${pvc.namespace}`
+      * Automatically replaced with the namespace of the `PersistentVolumeClaim` object being provisione
+
+The value of `csi.storage.k8s.io/provisioner-secret-name` also supports the following template variables which are automatically replaced by the `external-provisioner` at provision time:
+
+  * `${pvc.name}`
+    * Automatically replaced with the name of the `PersistentVolumeClaim` object being provisioned.
 
 If specified, the CSI `external-provisioner` will attempt to fetch the secret before provisioning and deletion.
 
-If no such secret exists in the Kubernetes API, or the provisioner is unable to fetch it, the provision or delete operation fails.
+If the entire namespace was deleted, including the secret needed for deletion, then no secret will be passed to the delete call. 
+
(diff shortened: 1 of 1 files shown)


book/src/secrets-and-credentials.md
```diff
@@ -48,6 +48,13 @@ The value of both parameters may be a literal or a template containing the follo
...
       * Automatically replaced with the name of the `PersistentVolume` object being provisioned at provision.
+    * `${pvc.namespace}`
+      * Automatically replaced with the namespace of the `PersistentVolumeClaim` object being provisione
+
+The value of `csi.storage.k8s.io/provisioner-secret-name` also supports the following template variables which are automatically replaced by the `external-provisioner` at provision time:
+
+  * `${pvc.name}`
+    * Automatically replaced with the name of the `PersistentVolumeClaim` object being provisioned.
 
 If specified, the CSI `external-provisioner` will attempt to fetch the secret before provisioning and deletion.
```
msau42: Can you also add a note here that for deletion, if the entire namespace was deleted (including the secret needed for deletion), then no secret will be passed to the delete call.  If the driver requires a deletion secret and this happens, then the volume and the PV may need to be manually cleaned up.

ggriffiths: Added a note for this :+1:

book/src/secrets-and-credentials.md
```diff
@@ -48,9 +48,20 @@ The value of both parameters may be a literal or a template containing the follo
...
 
+If the entire namespace was deleted, including the secret needed for deletion, then no secret will be passed to the
+delete call. If this happens when the driver requires a secret for deletion, then the volume and PV may need to be
+manually cleaned up. 
+
 If no such secret exists in the Kubernetes API, or the provisioner is unable to fetch it, the provision or delete operation fails.
```
msau42: This sentence also has to be updated :-)

ggriffiths: Updated this and moved the sentence about manually cleaning up to be after this.

PR #170, opened 2019-06-05
Doc update for CSI inline ephemeral volume support

vladimirvivien:
This is a document update for CSI inline ephemeral volume support.

```release-note
Support for CSI inline ephemeral volume
```


book/src/pod-inline-volume.md
```diff
@@ -0,0 +1,50 @@
+# Pod Inline Volume Support
+
+## Status
+
+Status | Min K8s Version | Max K8s Version 
+--|--|--
+Alpha | 1.15 | - 
+
+## Overview
+Traditionally, volumes that are backed by CSI drivers can only be used with a `PersistentVolume` and `PersistentVolumeClaim` object combination. This feature implements support that allows the nesting of volume declarations directly within pod specs for support of ephemeral storage using CSI drivers.  At runtime, nested inline volumes follow the ephemeral lifecycle of their associated pods where the driver handles all phases of volume operations as pods are created and destroyed.
```
msau42: 2nd sentence can maybe can be simplified to "This feature supports ephemeral storage use cases and allows CSI volumes to be specified directly in the pod specification."

book/src/pod-inline-volume.md
```diff
@@ -0,0 +1,50 @@
...
+              foo: bar
+```
+
+## Implementing inline ephemeral support
+
+Drivers must be modified (or implemented specifically) to support inline ephmeral workflows.  When Kubernetes encounters an inline CSI volume embedded in a pod spec, it treats that volume differently.  Mainly, the driver will only receive `NodePublish`, during the volume's mount phase, and `NodeUnpublish` when the pod is going away and the volume is unmounted.   To support inline, a driver must implement the followings:
+
+- Identity service
+- Node service
+- Specify `SkipAttach` using the [CSIDriver object](./csi-objects.md)
```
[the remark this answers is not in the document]
vladimirvivien: Well, not really necessary since Kubernetes will make the determination.  The KEP should be updated too.

book/src/pod-inline-volume.md
```diff
@@ -0,0 +1,50 @@
...
+  - If `SkipAttach` is not specified, Kubernetes will assume Volumes embedded in pod specs are to be used as inline.
+
+### Feature gates
+To use inline volume, Kubernetes binaries must start with the `CSIInlineVolume` feature gate enabled:
+```
+--feature-gates=CSIInlineVolume=true
+```
+
+### Example implementation
+See the CSI [Hostpath driver](https://github.com/kubernetes-csi/csi-driver-host-path) for an example of how to implement a driver with inline support.
```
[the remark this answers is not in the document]
vladimirvivien: I will also call the image driver from @kfox1111 . However, it's worth pointing out that Hostpath can support both mode.

book/src/pod-inline-volume.md
```diff
@@ -37,8 +37,10 @@ Drivers must be modified (or implemented specifically) to support inline ephmera
 
 - Identity service
 - Node service
-- Specify `SkipAttach` using the [CSIDriver object](./csi-objects.md)
-  - If `SkipAttach` is not specified, Kubernetes will assume Volumes embedded in pod specs are to be used as inline.
+- Specify `Mode` using the [CSIDriver object](./csi-objects.md) (to be supported in CSI inline Beta).
```
msau42: This was not implemented was it?

book/src/pod-inline-volume.md
```diff
@@ -0,0 +1,53 @@
+# Pod Inline Volume Support
```
msau42: Can this replace "ephemeral local volumes" page instead?

vladimirvivien: I replaced the link, take a look to see if that works for your suggestion.

msau42: Looks good can you rename the file to match the title of the page?

vladimirvivien: @msau42 Renamed.

msau42:
/approve
Can you squash the commits?

book/src/ephemeral-local-volumes.md
```diff
@@ -1,38 +1,49 @@
...
-1. Remote Persistent Volumes
-2. Local Persistent Volumes
-3. Local Ephemeral Volumes
+> See the [design document](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md) for futher information.
 
-The initial focus of Kubernetes CSI was Remote Persistent Volumes. However, the goal is for CSI to support all three types.
 
-This page documents how to create "Local Ephemeral Volumes" for Kubernetes using CSI.
+### Example of inline CSI pod spec
+A pod spec with an ephemeral inline CSI volume. Note that because the volume is expected to be ephemeral, the volumeHandle is not provided. Instead a CSI-generated ID will be submitted to the driver.
```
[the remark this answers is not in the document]
kfox1111: maybe just a "automatically generated id" or "kubelet generated id" rather then csi?

book/src/ephemeral-local-volumes.md
```diff
@@ -1,38 +1,49 @@
...
+          volumeAttributes:
+              foo: bar
+```
 
-A Local Ephemeral Volumes is a volume whose lifecycle is tied to the lifecycle of a single pod:
+## Implementing inline ephemeral support
 
-* The volume is "provisioned" (either empty or with some pre-populated data) when the pod is created.
-* The volume is deleted when the pod is terminated.
+Drivers must be modified (or implemented specifically) to support inline ephmeral workflows.  When Kubernetes encounters an inline CSI volume embedded in a pod spec, it treats that volume differently.  Mainly, the driver will only receive `NodePublish`, during the volume's mount phase, and `NodeUnpublish` when the pod is going away and the volume is unmounted.   To support inline, a driver must implement the followings:
```
pohly: s/ephmeral/ephemeral/

book/src/ephemeral-local-volumes.md
```diff
@@ -1,38 +1,49 @@
...
 
-[Kubernetes Secret Volumes](https://kubernetes.io/docs/concepts/storage/volumes/#secret) are a good example (non-CSI) of a local ephemeral volumes.
+- Identity service
+- Node service
```
pohly: As far as I understand it, there's more that is needed during deployment to make this work:
- configure `PodSecurityPolicy` (https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#specifying-allowed-inline-drivers-with-podsecuritypolicy)
- define driver mode (https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#driver-mode)

It would be useful to mention this here.

Note that the KEP is inconsistent regarding `PodSecurityPolicy`: the comment says that "Empty or nil indicates that all CSI drivers may be used." but the plain text then says "An empty value means no CSI drivers are allowed to be specified inline inside a pod spec." - which one is correct?

book/src/ephemeral-local-volumes.md
```diff
@@ -1,38 +1,49 @@
...
-  * This feature provides the CSI driver pod information at mount time. Many ephemeral volumes write some files at mount time. Often the data they write depends on the the pod they are operating on.
-* [Skip Attach](skip-attach.md)
-  * This instructs Kubernetes to skip any attach operation (`ControllerPublishVolume`) altogether. Local ephemeral volume drivers generally do not have or need a cluster control plane component.
-
-Features currently in development to improve Local Ephemeral Volume support:
-
-* Inline Volume Support
-  * Having to create a PV and a PVC for every ephemeral volume is onerous. Being able to specify a volume inside a pod definition (not currently possible for CSI drivers) will make that easier.
+### Example implementation
+- [CSI Hostpath driver](https://github.com/kubernetes-csi/csi-driver-host-path) - an example driver that supports either both PVC/PV and inline ephemeral volumes.
```
pohly: "that supports either PVC/PV or inline ephemeral volumes"?

book/src/ephemeral-local-volumes.md
kfox1111: PodSecurityPolicy is only required if you actually have a PodSecurityPolicy.

Driver mode is in the spec but not currently implemented. In the alpha, it just assumes it supports ephemeral if you try and use it in pod inline mode.

book/src/ephemeral-local-volumes.md
kfox1111: Yeah, thats kind of unclear.

The driver supports both modes, but can only be launched in one mode at a time.

vladimirvivien:
Wow, lots of new reviews!! Thanks. Will get them addressed.

book/src/ephemeral-local-volumes.md
msau42: We did not document driver mode here because it's currently unimplemented.

book/src/ephemeral-local-volumes.md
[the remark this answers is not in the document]
saad-ali: Maybe worth explaining how/when/by whom/why it is generated (this is an important difference between normal CSI volumes and ephemeral.

book/src/ephemeral-local-volumes.md
```diff
@@ -1,38 +1,49 @@
...
+  name: some-pod
+spec:
+  containers:
+    ...
+  volumes:
+      - name: vol
+        csi:
+          driver: inline.storage.kubernetes.io
+          volumeAttributes:
+              foo: bar
```
kfox1111: Not necessarily trying to promote the image driver, but the example from there may be more concrete an example and more easy to understand? This one is a little generic. Should we use that example?

```yaml
apiVersion: v1
kind: Pod
spec:
  containers:
  - image: nginx:1.13-alpine
    volumeMount:
    - name: data
      mountPath: /usr/share/nginx/html
  volumes:
  - name: data
    csi:
      driver: image.csi.k8s.io
      volumeAttributes:
          image: kfox1111/misc:test
```

pohly: I think that makes sense. But then we should also link to the documentation for the driver, either in a comment above `driver: ...` or in the introduction of the example. The latter would have the advantage that it can be a clickable link.

pohly:
@vladimirvivien: you need to rebase.

book/src/ephemeral-local-volumes.md
msau42: My only concern about this example is the reference to a private, unknown image. It may be better to refer to some publicly known image like busybox

book/src/ephemeral-local-volumes.md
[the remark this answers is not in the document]
vladimirvivien: Ok

book/src/ephemeral-local-volumes.md
vladimirvivien: Thanks for helping clarify this.

book/src/ephemeral-local-volumes.md
vladimirvivien: Will revise wording.

book/src/ephemeral-local-volumes.md
vladimirvivien: Seems like the consensus here it to keep the example as is using the Hostpath.

kfox1111: The only weird there is it shows an example where a regular container (executable binaries) is being shared out by a webserver. It may be confusing to the reader. "why would you put busybox there?"

Should we produce a container that could be used for this purpose? Something like:
kubernetes-csi/test-ephemeral-image

index.html:
Hello World!

Dockerfile:
FROM scratch
COPY index.html /index.html

?

kfox1111: > Seems like the consensus here it to keep the example as is using the Hostpath.

As is works for me.

HostPath is kind of confusing, as it raises the question, why not just emptyDir?

vladimirvivien:
@msau42 @kfox1111 @pohly  PTAL

book/src/ephemeral-local-volumes.md
kfox1111: "either both"

book/src/ephemeral-local-volumes.md
vladimirvivien: @kfox1111 please raise this point against the hostpath driver itself so that this documentation PR can keep moving.

pohly: The example doesn't really help someone who's new to the topic. For example, "inline.storage.kubernetes.io" looks like some magic string that always has to be there. It's not obvious that this the name of a fake CSI driver. Same with "foo" and "bar" - what is that about?

How about this:
```
volumes:
  - name: data
    csi:
      driver: some-csi-driver.example.com
      # volumeAttributes contains arbitrary key/value pairs,
      # valid options depend on the driver.
      volumeAttributes:
          dataSource: from-somewhere
```

book/src/ephemeral-local-volumes.md
adisky: i am trying to implement inline volume support for cinder csi driver, few queries here since mode is not supported and  `feature-gates=CSIInlineVolume=true` , is it assumed a driver is supporting both modes persistent+ephemeral??

adisky: One more thing i got from inline KEP, if requested volume is ephemeral, it will be notified in `NodePublishRequest.volume_context ` as `csi.storage.k8s.io/ephemeral: true`, it would be useful if this information is added in the doc for driver developers looking to support both persistent+ephemeral

adisky: ok, got it from here https://github.com/kubernetes-csi/docs/pull/170/files 
`csi.storage.k8s.io/ephemeral: true`  is not supported in v1.15, **so as of v1.15 at one time driver can be run in only one mode persistent or ephemeral,** could we specify this information in the doc?

vladimirvivien: > i am trying to implement inline volume support for cinder csi driver, few queries here since mode is not supported and `feature-gates=CSIInlineVolume=true` , is it assumed a driver is supporting both modes persistent+ephemeral??

That feature to support both is being worked on by @pohly for 1.16. Current behavior is to support one or the other.

vladimirvivien: > One more thing i got from inline KEP, if requested volume is ephemeral, it will be notified in `NodePublishRequest.volume_context ` as `csi.storage.k8s.io/ephemeral: true`, it would be useful if this information is added in the doc for driver developers looking to support both persistent+ephemeral

Thanks.   That is a good point, please open an issue against the github.com/kubernetes-csi/docs so this can be tracked.

msau42: For now let's only document what's possible as of 1.15.  we'll update this again once 1.16 is released

adisky: @msau42 agreed, but IMO as of 1.15 it should be explicitly specified in doc `that a driver can be support one mode only, feature to support both modes in parallel will be added in v1.16,`
 i have seen it is specified in the example section below of hostpath driver, but from there it is inferred as hostpath is developed in such a way, it is the limitation of the driver.

vladimirvivien:
@kfox1111 PTAL. Made the word changes.

book/src/ephemeral-local-volumes.md
pohly: I'll get to that soon. Let's merge this revision of the docs as-is.

PR #179, opened 2019-06-21
Fix table

j-griffith:
Add missing symbol for table building

```release-note
NONE
```


j-griffith:
@msau42 Yeah, to your point about a better way, I missed a line :)  I was thinking this didn't seem like it's going to scale well, don't have an immediate solution to propose other than pushing back down into the plugin docs.  Not many other options, particularly ones that are "better" but I'll think about it.

PR #181, opened 2019-07-08
Add note about storage classes to cloning doc

j-griffith:
Clarify the reasoning behind the storage class restriction for volume
cloning.

Address issue:  https://github.com/kubernetes-csi/external-provisioner/issues/308

```release-note
NONE
```

### DIFF
--- a/book/src/volume-cloning.md
+++ b/book/src/volume-cloning.md
@@ -12,6 +12,12 @@ A Clone is defined as a duplicate of an existing Kubernetes Volume.  For more in
 
 For details regarding the kuberentes API for volume cloning, please see [kubernetes concepts](https://kubernetes.io/docs/concepts/storage/volume-pvc-datasource/).
 
+## Cloning Restrictions
+
+Clone operations assume the cloning of a volume is being performed by the storage device, often referred to as `smart cloning`.  Implementations that transfer data (ie creating a container, mounting volumes and performing a dd) are considered `Populators` and not cloning implementations.  In addition to the expectation that the actual clone operation is performed by the storage device, cloning is also restricted to a single storage class.  In other words, the destination volume *must* be in the same Storage Class as the volume specified in the `dataSource` parameter.  This may seem unnecessary for devices or plugins that manage multiple storage classes, and while the check could instead be on the `Provisioner` field of a storage class, allowing this would break how storage classes work.
+
+One example of how this creates a problem is due to the way we use a storage class to define a file system type.  If for example a deployment includes a storage class for xfs and another for ext4, cloning from an xfs volume to an ext4 volume may work fine from the storage devices perspective, however a user ends up with an xfs formatted volume in the ext4 storage class.  There are other situations where the storage class defines specific characteristics for a volume that would likely be lost if cloning between them; block-mode, qos settings, credentials etcetera.
+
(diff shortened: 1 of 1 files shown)


book/src/volume-cloning.md
```diff
@@ -12,6 +12,12 @@ A Clone is defined as a duplicate of an existing Kubernetes Volume.  For more in
...
+
+One example of how this creates a problem is due to the way we use a storage class to define a file system type.  If for example a deployment includes a storage class for xfs and another for ext4, cloning from an xfs volume to an ext4 volume may work fine from the storage devices perspective, however a user ends up with an xfs formatted volume in the ext4 storage class.  There are other situations where the storage class defines specific characteristics for a volume that would likely be lost if cloning between them; block-mode, qos settings, credentials etcetera.
```
msau42: I think actually snapshots may have the fstype issue.  I think it will restore fine (and ignore fstype).  But when it comes to mount, the mount will fail.  cc @jingxu97 @xing-yang 

But I think with snapshots, it's possible to enforce fstype during restore.

j-griffith: @jingxu97 @xing-yang any thoughts on this?  I'm happy to update the PR based on your input.  Afraid I forgot this one was out here.

xing-yang: I think most of time restoring to a different storage class can work, but I don't know if we can say for sure.  Parameters are opaque.  We don't know how CSI drivers use them.

jingxu97: So my understanding is when restoring snapshot to a new volume, the file system on the restored volume will be the same as the volume which is used to take snapshot. If it is the case, could we have a field in VolumeSnapshotContent to indicate this fsType information. During restoring, controller could check whether the fsType matches and fails the operation if it does not match?

j-griffith: > So my understanding is when restoring snapshot to a new volume, the file system on the restored volume will be the same as the volume which is used to take snapshot. If it is the case, could we have a field in VolumeSnapshotContent to indicate this fsType information. During restoring, controller could check whether the fsType matches and fails the operation if it does not match?

IMO it's not whether we can or not, it's whether it's the right thing to do.  If I have a Storage Class the specifies that it uses XFS for example I think the expectation is that a Volume from that SC would in fact have an XFS file system on it.  If I import a snapshot to that SC that is in fact ext4, it seems to me that's a problem. Even if we can provide a hint to the mounter.  At that point I'd argue that we instead should revisit the limitations on the StorageClass (ie allow more than one type of FS per storage class) in our design.  Does that seem far out?

xing-yang: For snapshots, we can't reliably check the Storage Class of the original PVC as both may not exist any more.  We need to fix the dependency problem first so that a PVC should not be deleted when there are still snapshots on it.  Anyway I think we can handle snapshots in a different PR.

Regard this sentence - "There are other situations where the storage class defines specific characteristics for a volume that would likely be lost if cloning between them; block-mode, qos settings, credentials etcetera", block-mode is not set in storage class so it should not be mentioned in the context of storage class.  QoS settings and credentials, etc. are inside Parameters?  Those are opaque parameters that CO won't parse anyway, so I don't know if we need to talk about them here.

PR #184, opened 2019-07-18
Add note about what versions are supported

jsafrane:
We talked about this on our meeting yesterday. We should support all containers that include a supported Kubernetes version between their min/max versions.

From tables of individual containers, it seems we're going to support quite lot of branches (e.g. at least 3 external-provisioners), because we don't really set max k8s version unless it really breaks something and at the same time we open new 1.x branch every Kubernetes release.

cc @msau42 @pohly @kubernetes-csi/csi-misc

### DIFF
--- a/book/src/sidecar-containers.md
+++ b/book/src/sidecar-containers.md
@@ -57,6 +57,13 @@ stay as close to the recommended Kubernetes version as possible.
 For more details on which features are supported with which Kubernetes versions
 and their corresponding sidecars, please see each feature's individual page.
 
+### Support
+
+Kubernetes community supports CSI sidecar container versions that cover at
+least one
+[supported Kubernetes version](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions)
+between their minimum and maximum versions.
+
 ### Alpha Features
 
 It is also important to note that alpha features are subject to break or be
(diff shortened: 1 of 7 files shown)


book/src/sidecar-containers.md
```diff
@@ -57,6 +57,13 @@ stay as close to the recommended Kubernetes version as possible.
 For more details on which features are supported with which Kubernetes versions
 and their corresponding sidecars, please see each feature's individual page.
 
+### Support
+
+Kubernetes community supports CSI sidecar container versions that cover at
+least one
+[supported Kubernetes version](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions)
+between their minimum and maximum versions.
```
pohly: But we also stop supporting old versions that have been completely superseded by a newer version.

For example, we currently still support external-provisioner release-1.0 because it is needed for Kubernetes 1.13 and that version is still supported. But once 1.13 is no longer supported, we can stop supporting release-1.0 even though it would also work on more recent Kubernetes. Anyone running on Kubernetes > 1.13 then can and should update to a new release.

I still think we should have either an explicit "supported: yes/no" column in the version table (https://kubernetes-csi.github.io/docs/external-provisioner.html) or a second table below with historic, unsupported releases.

Is Kubernetes 1.10 still supported? I don't know of the top of my head, so even if I know all of our rules for "supported or not supported", I can't quickly whether release-0.4 is still supported.

PR #197, opened 2019-08-22
update docs: support new features & update version

zhucan:
1. support volume expansion 
2. support raw block volume
3. support volume cloning
4. update csi plugin version to v2.2.0


zhucan:
@saad-ali

PR #209, opened 2019-09-13
ephemeral volumes in Kubernetes 1.16

pohly:
The CSIDriverInfo object was extended and the feature changed from
alpha to beta.

### DIFF
--- a/book/src/csi-driver-object.md
+++ b/book/src/csi-driver-object.md
@@ -26,6 +26,9 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: true
+  volumeLifecycleModes: # added in Kubernetes 1.16
+  - Persistent
+  - Ephemeral
 ```
 
 There are three important fields:
@@ -45,6 +48,13 @@ There are three important fields:
     * `"csi.storage.k8s.io/pod.namespace": pod.Namespace`
     * `"csi.storage.k8s.io/pod.uid": string(pod.UID)`
   * For more information see [Pod Info on Mount](pod-info.md).
+* `volumeLifecycleModes`
+  * This field was added in Kubernetes 1.16 and cannot be set when using an older Kubernetes release.
+  * It informs Kubernetes about the volume modes that are supported by the driver.
+    This ensures that the driver [is not used incorrectly](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#support-for-inline-csi-volumes) by users.
+    The default is `Persistent`, which is the normal PVC/PV mechanism. `Ephemeral` enables
+    [inline ephemeral volumes](ephemeral-local-volumes.md) in addition (when both
+    are listed) or instead of normal volumes (when it is the only entry in the list).
 
 ## What creates the CSIDriver object?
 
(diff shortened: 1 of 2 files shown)


pohly:
This assumes that csi-driver-host-path v1.2.0 gets released including https://github.com/kubernetes-csi/csi-driver-host-path/pull/97.

book/src/csi-driver-object.md
```diff
@@ -26,6 +26,9 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: true
+  volumeLifecycleModes:
```
msau42: Add a comment here that it's added in 1.16

book/src/csi-driver-object.md
```diff
@@ -45,6 +48,12 @@ There are three important fields:
     * `"csi.storage.k8s.io/pod.namespace": pod.Namespace`
     * `"csi.storage.k8s.io/pod.uid": string(pod.UID)`
   * For more information see [Pod Info on Mount](pod-info.md).
+* `volumeLifecycleModes`
+  * This field was added in Kubernetes 1.16 and must not be set when using an older Kubernetes release.
```
msau42: must => cannot

book/src/csi-driver-object.md
```diff
@@ -45,6 +48,12 @@ There are three important fields:
...
+  * It informs Kubernetes about the volume modes that are supported by the driver.
```
msau42: What does Kubernetes do with that information?

book/src/ephemeral-local-volumes.md
```diff
@@ -5,6 +5,7 @@
 Status | Min K8s Version | Max K8s Version 
 --|--|--
 Alpha | 1.15 | -
```
msau42: Alpha max is 1.15

book/src/csi-driver-object.md
pohly: Done.

book/src/csi-driver-object.md
pohly: I was thinking "must not be set in the .yaml file because it's invalid", but "cannot" is simpler. Changed.

book/src/csi-driver-object.md
pohly: Added: "This ensures that the driver is not used incorrectly by Kubernetes."

book/src/ephemeral-local-volumes.md
pohly: Done.

book/src/csi-driver-object.md
msau42: Can you be more specific about what enforcement or behavior changes result from setting the fields? Or link to the inline page that describes that?

book/src/ephemeral-local-volumes.md
```diff
@@ -38,12 +39,20 @@ Drivers must be modified (or implemented specifically) to support inline ephemer
 - Identity service
 - Node service
 
+Kubernetes 1.16 only allows using a CSI driver for an inline volume if
+its [`CSIDriverInfo`](csi-driver-object.md) object explicitly declares
+that the driver supports that kind of usage in its
+`volumeLifecycleModes` field. This is a safeguard against accidentally
+using a driver the wrong way.
```
pohly: @msau42: this is basically the same comment as in csi-driver-object.md ("used incorrectly by Kubernetes").

Should I expand on that in ephemeral-local-volumes.md, in  csi-driver-object.md, or link to https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#support-for-inline-csi-volumes?

I'm leaning towards keeping the current text and just turning it into a link to the KEP, but I'm also fine with everything else.

msau42: I would just link to the KEP for now.  I bring this up because there is an ask to remove this field if it's not really providing anything other than a nicer error message when someone tries to misuse a driver: https://github.com/kubernetes/kubernetes/issues/82507

book/src/csi-driver-object.md
```diff
@@ -45,6 +48,13 @@ There are three important fields:
     * `"csi.storage.k8s.io/pod.namespace": pod.Namespace`
     * `"csi.storage.k8s.io/pod.uid": string(pod.UID)`
   * For more information see [Pod Info on Mount](pod-info.md).
+* `volumeLifecycleModes`
+  * This field was added in Kubernetes 1.16 and cannot be set when using an older Kubernetes release.
+  * It informs Kubernetes about the volume modes that are supported by the driver.
+    This ensures that the driver [is not used incorrectly](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md#support-for-inline-csi-volumes) by Kubernetes.
```
msau42: "by Kubernetes" or "by users"?

pohly: "by Kubernetes on behalf of a user". That's becoming complicated - perhaps the sentence should simply end with "is not used incorrectly".

msau42: I think "by users" is fine

msau42: "by Kubernetes" implies to me that we're protecting against Kubernetes bugs, which I don't think is the case here

pohly: Changed to "by users".

PR #217, opened 2019-10-08
Consolidation of driver feature columns

### LINKED ISSUE #216: Driver Feature table is a bit out of control
Keeping a table of [production plugins and the features they implement](https://github.com/kubernetes-csi/docs/blob/master/book/src/drivers.md#production-drivers) is handy, but as the project has grown, along with the number of plugins and features;  this simple table has become a bit difficult to view and even more difficult to update and keep accurate.

I don't have a solution to reformat what's in place, but I wonder if instead of trying to keep this up to date for features, replace the features columns with a link to the plugins documentation?  Said link could be a plugin doc (we could provide a template for plugin authors to use) and we could host that in this repo under sub-sections (sub-section for each plugin).

The downside of this is that users like to have the matrix in front of them to compare things, so this doesn't help with that.  It does however clean things up in terms of listing available plugins and their features; it provides the information, just no easy way to compare plugins.

Opening this issue to propose the template idea or to solicit input for a better suggestion if someone has one.

wnxn:
Signed-off-by: Xin Wang <wileywang@yunify.com>

Fixes #216 

```release-note
NONE
```


j-griffith:
@wnxn Thanks for putting this together, I think it's an improvement in readability.  This markdown is still pretty brutal, but we can keep iterating on ideas IMO.

At first I wasn't sure how I felt about "no" under extra features, but maybe putting "no" or "none" there will prompt folks to update if needed.  We should probably come up with a reference for what "extra features" are though to along with this if we're not going to use separate pages; what do you think?

wnxn:
@j-griffith Thanks for your review. I agree with this markdown is pretty brutal and we can keep iterating. 

> We should probably come up with a reference for what "extra features" are though to along with this if we're not going to use separate pages.

Did you mean to add some notes or guides for people to fill in "extra features"?  Actually, I added notes in this PR. Now I modify it.
"Extra Features allows to be filled in Raw Block, Snapshot, Volume Expansion and Volume Cloning. If driver does not implemented any Extra Features, please fill in none". Can we add this guide behind NOTE? 

> NOTE: If you would like your driver to be added to this table, please open a pull request in [this repo](https://github.com/kubernetes-csi/docs/pulls) updating this file. Extra Features allows to be filled in Raw Block, Snapshot, Volume Expansion and Volume Cloning. If driver does not implemented any Extra Features, please fill in none.


I am also uncertain about some fields, `Other Features` or `Extra Features`, `no` or `none`. Which is better?

book/src/drivers.md
```diff
@@ -1,61 +1,61 @@
...
-[YanRongYun](http://www.yanrongyun.com/) | ? | v1.0 | A Container Storage Interface (CSI) Driver for YanRong YRCloudFile Storage  | Persistent | Read/Write Multiple Pods | Yes | No | No | ? | ?
+Name | CSI Driver Name | Compatible with CSI Version(s) | Description | Persistence (Beyond Pod Lifetime) | Supported Access Modes | Dynamic Provisioning | Other Features
+-----|-----------------|--------------------------------|-------------|-----------------------------------|------------------------|----------------------|--------
+[Alicloud Disk](https://github.com/AliyunContainerService/csi-plugin) | `diskplugin.csi.alibabacloud.com` | v1.0 | A Container Storage Interface (CSI) Driver for Alicloud Disk | Persistent | Read/Write Single Pod | Yes | Raw Block, Snapshot
+[Alicloud NAS](https://github.com/AliyunContainerService/csi-plugin) | `nasplugin.csi.alibabacloud.com` | v1.0 | A Container Storage Interface (CSI) Driver for Alicloud Network Attached Storage (NAS) | Persistent | Read/Write Multiple Pods | No | No
```
msau42: instead of "no" for other features, can we just leave it blank?

wnxn: Done.
I also update NOTE to explain how to fill in Other Features. Would you please take a look and give me some advice?

wnxn:
@j-griffith @msau42  PTAL

book/src/drivers.md
```diff
@@ -1,61 +1,61 @@
 # Drivers
 The following are a set of CSI driver which can be used with Kubernetes:
 
-> NOTE: If you would like your driver to be added to this table, please open a pull request in [this repo](https://github.com/kubernetes-csi/docs/pulls) updating this file.
+> NOTE: If you would like your driver to be added to this table, please open a pull request in [this repo](https://github.com/kubernetes-csi/docs/pulls) updating this file. Other Features allows to be filled in Raw Block, Snapshot, Volume Expansion and Volume Cloning. If driver did not implement any Other Features, please leave it blank.
```
msau42: typo: allowed

Also can we shorten feature names by dropping the "Volume"? So just "Expansion" and "Cloning"

wnxn: > typo: allowed
> 
> Also can we shorten feature names by dropping the "Volume"? So just "Expansion" and "Cloning"

Done

PR #241, opened 2019-12-04
Update volume expansion status for 1.14 through 1.16

timoreimann:


book/src/volume-expansion.md
```diff
@@ -4,7 +4,8 @@
 
 Status | Min K8s Version | Max K8s Version | external-resizer Version
 --|--|--|--
-Alpha | 1.14 | - | 0.1
+Alpha | 1.14 | 1.15 | 0.2
```
timoreimann: Note to reviewers: I'm assuming here that v0.2 is compatible with both 1.14 and 1.15.

timoreimann:
@msau42 @gnufied updated the section about enabling volume expansion as well.

PR #243, opened 2019-12-09
Update change log for CSI spec 1.1.0 and 1.2.0

timoreimann:
I assumed that both CSI releases went GA with their respective Kubernetes versions.

Also assumed Volume capacity usage metrics went GA with 1.15 in order to group it under the GA header.

Please let me know if either is incorrect.

```release-note
NONE
```

### DIFF
--- a/book/src/kubernetes-changelog.md
+++ b/book/src/kubernetes-changelog.md
@@ -14,6 +14,7 @@ details on individual features, visit the [Features section](features.md).
 
 ### Features
 * GA
+    * Support for [CSI spec 1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0).
     * Volume topology
     * Volume limits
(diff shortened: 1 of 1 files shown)


timoreimann:
Not sure about the release notes either; I noted down `NONE` for now.

book/src/kubernetes-changelog.md
```diff
@@ -14,6 +14,7 @@ details on individual features, visit the [Features section](features.md).
 
 ### Features
 * GA
+    * Support for [CSI spec 1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0).
```
msau42: We may want to clarify exactly which parts of the new spec we support, since k8s currently doesn't implement everything. Cc @gnufied for suggestions

PR #255, opened 2019-12-31
modify the error url of e2e-tests

timyinshi:
/kind bug
/sig docs

the error url is:
https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md

the right url is:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md

### DIFF
--- a/book/src/_book/Troubleshooting.html
+++ b/book/src/_book/Troubleshooting.html
@@ -401,7 +401,7 @@ E0306 16:34:50.992034       1 reflector.go:205] github.com/kubernetes-csi/extern
 happen that they become incompatible with each other. If the
  issues above above have been ruled out, <a href="https://github.com/kubernetes/community/tree/master/sig-storage" target="_blank">contact the sig-storage
 team</a> and/or
-<a href="https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md#local-clusters" target="_blank">run the e2e test</a>:</p>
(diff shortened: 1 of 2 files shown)


book/src/_book/Troubleshooting.html
```diff
@@ -401,7 +401,7 @@ <h3 id="problems-with-the-external-components">Problems with the external compon
 happen that they become incompatible with each other. If the
  issues above above have been ruled out, <a href="https://github.com/kubernetes/community/tree/master/sig-storage" target="_blank">contact the sig-storage
 team</a> and/or
-<a href="https://github.com/kubernetes/community/blob/master/contributors/devel/e2e-tests.md#local-clusters" target="_blank">run the e2e test</a>:</p>
+<a href="https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/e2e-tests.md#local-clusters" target="_blank">run the e2e test</a>:</p>
```
msau42: I'm not sure what this "_book" directory is used for. It hasn't been updated in a very long time. @lpabon do you know?

pohly: This might have been used at the time when the .html files had to be updated as part of the PR for this repo. But I am not sure because `make all` puts generated .html files directly into `docs`. I removed them in https://github.com/kubernetes-csi/docs/pull/82/files when switching to HTML creation via TravisCI.

But then merging some PR that was based on a revision before that change seems to have brought back the directory.

It should be safe to remove these files => https://github.com/kubernetes-csi/docs/pull/256

PR #258, opened 2020-01-07
Update external-attacher sidecar version

jsafrane:
cc @msau42

### DIFF
--- a/book/src/external-attacher.md
+++ b/book/src/external-attacher.md
@@ -8,6 +8,7 @@
 
 Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min K8s Version | Max K8s Version | Recommended K8s Version |
 --|--|--|--|--|--|--|--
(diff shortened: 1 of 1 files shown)


book/src/external-attacher.md
```diff
@@ -8,6 +8,7 @@
 
 Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min K8s Version | Max K8s Version | Recommended K8s Version |
 --|--|--|--|--|--|--|--
+[external-attacher v2.1.0](https://github.com/kubernetes-csi/external-attacher/releases/tag/v2.1.0) | [release-2.1](https://github.com/kubernetes-csi/external-attacher/tree/release-2.1) | [v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | - | quay.io/k8scsi/csi-attacher:v2.1.0 | v1.14 | - | v1.15
```
msau42: Do we want recommended to be 1.17 for csimigration beta?

jsafrane: There was no change in the attacher related to CSI migration in this release. It looks odd to recommend 1.17 for v2.1, while the code is the same as in v2.0.

msau42: Should we replace the 2.0.0 line then?

jsafrane: That looks even worse, since 2.0.0 was released when 1.17 did not exist yet :-)

Ok, I am recommending 1.17 in 2.1.0, as the migration was more tested there.

PR #266, opened 2020-01-22
Update ceph-csi feature support

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>


humblec:
@saad-ali @msau42 @rootfs @pohly @lpabon ptal.

book/src/drivers.md
```diff
@@ -15,8 +15,8 @@ Name | CSI Driver Name | Compatible with CSI Version(s) | Description | Persiste
...
 [Azure disk](https://github.com/kubernetes-sigs/azuredisk-csi-driver) | `disk.csi.azure.com` | v0.3, v1.0 | A Container Storage Interface (CSI) Driver for Azure disk | Persistent | Read/Write Single Pod | Yes | 
 [Azure file](https://github.com/kubernetes-sigs/azurefile-csi-driver) | `file.csi.azure.com` | v0.3, v1.0 | A Container Storage Interface (CSI) Driver for Azure file | Persistent | Read/Write Multiple Pods | Yes | 
-[CephFS](https://github.com/ceph/ceph-csi) | `cephfs.csi.ceph.com` | v0.3, v1.0.0, v1.1.0 | A Container Storage Interface (CSI) Driver for CephFS | Persistent | Read/Write Multiple Pods | Yes | 
-[Ceph RBD](https://github.com/ceph/ceph-csi) | `rbd.csi.ceph.com` | v0.3, v1.0.0, v1.1.0 | A Container Storage Interface (CSI)  Driver for Ceph RBD | Persistent | Read/Write Single Pod | Yes | Raw Block, Snapshot
+[CephFS](https://github.com/ceph/ceph-csi) | `cephfs.csi.ceph.com` | v0.3, v1.0.0, v1.1.0, v1.2.0, v2.0.0 | A Container Storage Interface (CSI) Driver for CephFS | Persistent | Read/Write Multiple Pods | Yes | Resize, multiarch
```
Madhu-1: @humblec v2.0.0 should not be added here, the column is for CSI spec version.

Madhu-1:
@humblec commit message and PR title need to be updated

humblec:
> @humblec commit message and PR title need to be updated

PTAL

book/src/drivers.md
```diff
@@ -15,8 +15,8 @@ Name | CSI Driver Name | Compatible with CSI Version(s) | Description | Persiste
...
 [Azure disk](https://github.com/kubernetes-sigs/azuredisk-csi-driver) | `disk.csi.azure.com` | v0.3, v1.0 | A Container Storage Interface (CSI) Driver for Azure disk | Persistent | Read/Write Single Pod | Yes | 
 [Azure file](https://github.com/kubernetes-sigs/azurefile-csi-driver) | `file.csi.azure.com` | v0.3, v1.0 | A Container Storage Interface (CSI) Driver for Azure file | Persistent | Read/Write Multiple Pods | Yes | 
-[CephFS](https://github.com/ceph/ceph-csi) | `cephfs.csi.ceph.com` | v0.3, v1.0.0, v1.1.0 | A Container Storage Interface (CSI) Driver for CephFS | Persistent | Read/Write Multiple Pods | Yes | 
-[Ceph RBD](https://github.com/ceph/ceph-csi) | `rbd.csi.ceph.com` | v0.3, v1.0.0, v1.1.0 | A Container Storage Interface (CSI)  Driver for Ceph RBD | Persistent | Read/Write Single Pod | Yes | Raw Block, Snapshot
+[CephFS](https://github.com/ceph/ceph-csi) | `cephfs.csi.ceph.com` | v0.3, v1.0.0, v1.1.0 | A Container Storage Interface (CSI) Driver for CephFS | Persistent | Read/Write Multiple Pods | Yes | Resize, multiarch
```
Madhu-1: %s/Resize/Expansion

%s/multiarch/Multi Architecture ?

msau42: The additional features column is intended for CSI features. I think multiarch probably doesn't need to be called out here, and instead should be called out in the driver repo.

xing-yang: What does "multi architecture" mean?

saad-ali: +1 let's remove `multiarch` unclear what that means

jsafrane: In similar fashion, CSI does not support encryption, i.e. it does not have dedicated fields for it in any of its messages. It can be implemented by the driver, for example using StorageClass parameters, but these are opaque to CSI. I believe most of the drivers on this page support some sort of encryption under the hood.

humblec: > The additional features column is intended for CSI features. I think multiarch probably doesn't need to be called out here, and instead should be called out in the driver repo.

Sure, removed `multiarch` from the features list @msau42 @xing-yang @saad-ali . @xing-yang last month we released v2.0.0 of ceph-csi which support `arm64` arch of the driver, thats what I meant with that change. PTAL.

humblec: > In similar fashion, CSI does not support encryption, i.e. it does not have dedicated fields for it in any of its messages. It can be implemented by the driver, for example using StorageClass parameters, but these are opaque to CSI. I believe most of the drivers on this page support some sort of encryption under the hood.

sure @jsafrane , some how I had a feel that, its better to list storage specific functionalities which driver support in this doc, but you are correct, that could cause confusion and will have an inconsistent list of features for different drivers. Removed `encryption` too. thanks. PTAL.

PR #290, opened 2020-04-09
Update docs for 1.18

msau42:


book/src/external-attacher.md
```diff
@@ -8,7 +8,7 @@
 
 Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min K8s Version | Max K8s Version | Recommended K8s Version |
 --|--|--|--|--|--|--|--
-[external-attacher v2.1.0](https://github.com/kubernetes-csi/external-attacher/releases/tag/v2.1.0) | [release-2.1](https://github.com/kubernetes-csi/external-attacher/tree/release-2.1) | [v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | - | quay.io/k8scsi/csi-attacher:v2.1.0 | v1.14 | - | v1.17
+[external-attacher v2.2.0](https://github.com/kubernetes-csi/external-attacher/releases/tag/v2.2.0) | [release-2.2](https://github.com/kubernetes-csi/external-attacher/tree/release-2.2) | [v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | - | quay.io/k8scsi/csi-attacher:v2.2.0 | v1.14 | - | v1.17
```
msau42, line 11: @jsafrane the attacher release notes says recommended 1.18. Is that correct? What 1.18 feature does the latest attacher use?

book/src/external-provisioner.md
```diff
@@ -8,6 +8,7 @@
 
 Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min K8s Version | Max K8s Version | Recommended K8s Version
 --|--|--|--|--|--|--|--
+[external-provisioner v1.6.0](https://github.com/kubernetes-csi/external-provisioner/releases/tag/v1.6.0) | [release-1.6](https://github.com/kubernetes-csi/external-provisioner/tree/release-1.6) | [v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | - | quay.io/k8scsi/csi-provisioner:v1.6.0 | v1.13 | - | v1.18
```
xing-yang, line 11: Is the min k8s version still 1.13?

book/src/external-snapshotter.md
```diff
@@ -10,7 +10,7 @@
 
 Latest stable release | Branch | Min CSI Version | Max CSI Version | Container Image | Min K8s Version | Max K8s Version | Recommended K8s Version
 --|--|--|--|--|--|--|--
-[external-snapshotter v2.0.1](https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v2.0.1) | [release-2.0](https://github.com/kubernetes-csi/external-snapshotter/tree/release-2.0) | [v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | - | quay.io/k8scsi/csi-snapshotter:v2.0.1 | v1.17 | - | v1.17
+[external-snapshotter v2.1.0](https://github.com/kubernetes-csi/external-snapshotter/releases/tag/v2.1.0) | [release-2.1](https://github.com/kubernetes-csi/external-snapshotter/tree/release-2.1) | [v1.0.0](https://github.com/container-storage-interface/spec/releases/tag/v1.0.0) | - | quay.io/k8scsi/csi-snapshotter:v2.1.0 | v1.17 | - | v1.17
```
xing-yang, line 13: Should the "recommended k8s version" be renamed to "recommended min  k8s version" instead?  I don't see any reason why it works better in 1.17 than 1.18?

book/src/external-provisioner.md
msau42: Yes, if you don't use any of the optional features added after 1.13, then provisioner still works

book/src/external-snapshotter.md
msau42: Yes "recommended min version" is more accurate. Or maybe "fully-featured min version"?

xing-yang: "fully-featured" sounds a little weird:).  Maybe "recommended min version" is better.

msau42: Thinking about it some more, I think saying recommended min gets tricky around alpha features which may break in future releases.

xing-yang: That also applies to "recommended version"?

xing-yang: I guess if we say 1.17 is the recommended version, does that suggest we have not tested with 1.18 or something does not work with 1.18?

msau42: It's tricky. We test the recommended version+ until we release another sidecar version with a higher recommended version.

For example, external-attacher 2.0 has recommended 1.15. So we run tests using 2.0 on 1.15 and 1.16.  But 2.1 recommended is 1.17. So we run tests using 2.1 using 1.17 (and 1.18)

msau42: Actually...... I take that back.

We do test all future K8s versions. The hostpath-driver deployment X reflects the recommend versions.

And then we do skew testing for all future releases, ie deploy 1.16 on k8s 1.16, 1.17, 1.18.  deploy 1.17 on k8s 1.17, 1.18. deploy 1.18 on k8s 1.18

msau42: https://k8s-testgrid.appspot.com/sig-storage-csi-ci

msau42:
Updated to include links to definitions of min/max/recommended

Issue #305, opened 2020-04-28
Cloning Docs Need to be updated


saad-ali:
https://kubernetes-csi.github.io/docs/volume-cloning.html

Volume Cloning for Kubernetes went GA in Kubernetes 1.18. The docs need to be updated to reflect that.

j-griffith:
I'll get this submitted ASAP

saad-ali:
Fixed in #306

PR #306, opened 2020-04-28
Mark Cloning as GA

j-griffith:
Cloning was promoted to GA in the 1.18 release.  Update the CSI docs to
reflect that.  No functional changes were made to the
external-provisioner since beta so everything else in the docs should be
fine as is.

ISSUE: #305 

```release-note
Docs updated to reflect GA status of Cloning in 1.18 Kubernetes release
```

### DIFF
--- a/book/src/volume-cloning.md
+++ b/book/src/volume-cloning.md
@@ -5,7 +5,8 @@
 Status | Min k8s Version | Max k8s version | external-provisioner Version
 --|--|--|--
 Alpha | 1.15 | 1.15 | 1.3
-Beta | 1.16 | - | 1.4
+Beta | 1.16 | 1.17 | 1.4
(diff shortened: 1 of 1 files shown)


book/src/volume-cloning.md
```diff
@@ -5,7 +5,8 @@
 Status | Min k8s Version | Max k8s version | external-provisioner Version
 --|--|--|--
 Alpha | 1.15 | 1.15 | 1.3
-Beta | 1.16 | - | 1.4
+Beta | 1.16 | 1.17 | 1.4
+GA   | 1.18 | - | 1.4
```
msau42, line 9: We marked cloning as GA in provisioner v1.6.0, because it has a couple of fixes and the finalizer thing.

PR #350, opened 2020-07-20
Specify that block volumes are GA

ialidzhikov:
In https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/531#discussion_r457141148 @bertinatto noticed that the block volume feature is GA starting from v1.18 (see https://github.com/kubernetes/kubernetes/pull/88673/).

### DIFF
--- a/book/src/raw-block.md
+++ b/book/src/raw-block.md
@@ -6,7 +6,8 @@ Status | Min K8s Version | Max K8s Version | external-provisioner Version | exte
 --|--|--|--|--
 Alpha | 1.11 | 1.13 | 0.4 | 0.4
 Alpha | 1.13 | 1.13 | 1.0 | 1.0
-Beta | 1.14 | - | 1.1+ | 1.1+
+Beta | 1.14 | 1.17 | 1.1+ | 1.1+
+GA | 1.18 | - | 1.1+ | 1.1+
 
 ## Overview
 


book/src/raw-block.md
```diff
@@ -6,7 +6,8 @@ Status | Min K8s Version | Max K8s Version | external-provisioner Version | exte
 --|--|--|--|--
 Alpha | 1.11 | 1.13 | 0.4 | 0.4
 Alpha | 1.13 | 1.13 | 1.0 | 1.0
-Beta | 1.14 | - | 1.1+ | 1.1+
+Beta | 1.14 | 1.18 | 1.1+ | 1.1+
```
msau42, line 9: I think max should be 1.17

book/src/raw-block.md
```diff
@@ -6,7 +6,8 @@ Status | Min K8s Version | Max K8s Version | external-provisioner Version | exte
...
+GA | 1.18 | - | - | -
```
msau42, line 10: @jsafrane what version of external provisioner do you want to put here. 1.6? And attacher? Does it have any logic related to block volumes?

jsafrane: * Attacher supports attaching block volumes since v1.1.0 and there was not any change in the code since then.
* The main provisioner code supports block since v1.1.0 and again, no changes/bugfixes there. Block support was gradually added to cloning and snapshots. I would say that v1.1+ is the most appropriate, as cloning and snapshots have their own feature.

book/src/raw-block.md
ialidzhikov: Yep, you are right. Thanks for the catch.

book/src/raw-block.md
ialidzhikov: Thanks for bringing this up and thanks for the answer. Actually I was wondering which versions to pick, trying to check the changelog of the components but I couldn't find something related there. So now I updated the PR to `1.1+` also for the GA status..

jsafrane:
@xing-yang @msau42, can you please approve?

PR #385, opened 2020-11-13
Move CSIVolumeFSGroupPolicy to beta

huffmanca:
The CSIVolumeFSGroupPolicy feature is moving to beta in k8s 1.20. This PR updates the documentation accordingly.

PRs:
kubernetes/kubernetes#95739

Issues:
kubernetes/enhancements#1682

```release-note
The CSIVolumeFSGroupPolicy field has moved to beta, allowing storage.k8s.io/CSIDrivers to specify if they support modifying volume ownership and permissions when volumes are being mounted.
```

### DIFF
--- a/book/src/csi-driver-object.md
+++ b/book/src/csi-driver-object.md
@@ -30,7 +30,7 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: true
-  fsGroupPolicy: File # added in Kubernetes 1.19, this field is alpha
+  fsGroupPolicy: File # added in Kubernetes 1.19, this field is beta as of Kubernetes 1.20
   volumeLifecycleModes: # added in Kubernetes 1.16, this field is beta
     - Persistent
     - Ephemeral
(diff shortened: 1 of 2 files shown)


book/src/csi-driver-object.md
```diff
@@ -27,7 +27,7 @@ metadata:
 spec:
   attachRequired: true
   podInfoOnMount: true
-  fsGroupPolicy: File # added in Kubernetes 1.19, this field is alpha
+  fsGroupPolicy: File # added in Kubernetes 1.19, this field is beta
```
msau42, line 30: I think we should clarify it's alpha in 1.19, beta in 1.20

msau42 (commented):
/approve

book/src/csi-driver-object.md
```diff
@@ -61,7 +61,7 @@ These are the important fields:
   - For more information see [Pod Info on Mount](pod-info.md).
 - `fsGroupPolicy`
   - This field was added in Kubernetes 1.19 and cannot be set when using an older Kubernetes release.
-  - This field is alpha.
+  - This field is beta.
```
msau42, line 64: beta in 1.20

huffmanca: I've added it here, too; however, should we do this with the other fields (in a separate PR)?

msau42: Yes I think updating the other fields makes sense

PR #388, opened 2020-11-26
Improvement in snapshot-restore feature.md

Kartik494:
Added Invalid snapshot yaml file in Test snapshot feature

### DIFF
--- a/book/src/snapshot-restore-feature.md
+++ b/book/src/snapshot-restore-feature.md
@@ -146,6 +146,10 @@ Create a _PVC_ from a _VolumeSnapshot_:
 kuberctl create -f restore.yaml
 ```
 
+To create an invalid VolumeSnapshot:
+```
+kubectl create -f invalid-snapshot.yaml
+```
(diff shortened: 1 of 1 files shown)


book/src/snapshot-restore-feature.md
```diff
@@ -146,6 +146,10 @@ Create a _PVC_ from a _VolumeSnapshot_:
 kuberctl create -f restore.yaml
 ```
 
+To create an invalid VolumeSnapshot:
```
xing-yang, line 149: We don't want to encourage user to create an invalid volume snapshot.  This is an example to test the validation webhook feature.  I don't think we need to add it here.  It will actually add confusion.

Kartik494: @xing-yang sure , my intention was here to brief user about the example yaml for an invalid  volume snapshot , i am totally agreed with you if it is creating confusion, then this change should not be implemented here

Kartik494: /close

PR #391, opened 2020-11-30
fix csi spec compatibility support for vsphere driver

divyenpatel:
vSphere CSI driver is compatible with CSI Spec v1.0. We misunderstood this field and thought it is for the current driver version.

cc: @xing-yang


xing-yang:
/release-note-none

Need rebase.

PR #272, opened 2020-02-04
CSI Extension for Ephemeral Volume Support

lpabon:
```release-note
NONE
```

PR #397, opened 2020-12-07
CSI Extension for Ephemeral Volume Support, generic ephemeral

pohly:
This finishes PR #272 and adds information about generic ephemeral inline volumes to avoid confusion.


book/src/ephemeral-local-volumes.md
```diff
@@ -2,15 +2,44 @@
 
 ## Status
```
msau42, line 3: should we remove "local" from the page name?

book/src/ephemeral-local-volumes.md
```diff
@@ -2,15 +2,44 @@
...
+
+### Generic Ephemeral Inline Volumes
+
+Status | Min K8s Version | Max K8s Version
+--|--|--
+Alpha | 1.19 | -
 
 ## Overview
-Traditionally, volumes that are backed by CSI drivers can only be used with a `PersistentVolume` and `PersistentVolumeClaim` object combination. This feature supports ephemeral storage use cases and allows CSI volumes to be specified directly in the pod specification.  At runtime, nested inline volumes follow the ephemeral lifecycle of their associated pods where the driver handles all phases of volume operations as pods are created and destroyed.
+Traditionally, volumes that are backed by CSI drivers can only be used with a `PersistentVolume` and `PersistentVolumeClaim` object combination. The "CSI Ephemeral Inline volumes" feature allows CSI volumes to be specified directly in the pod specification for ephemeral use cases.  At runtime, nested inline volumes follow the ephemeral lifecycle of their associated pods where the driver handles all phases of volume operations as pods are created and destroyed.
```
msau42, line 19: What if we start off the overview and introduce both options, then go into more detail into each? Something like:

"Traditionally... combination. Two different Kubernetes features allow volumes to follow the Pod's lifecycle: CSI Ephemeral volumes and Generic ephemeral volumes..."

book/src/ephemeral-local-volumes.md
```diff
@@ -2,15 +2,44 @@
...
+"*generic* ephemeral inline volumes".
+
+> See the [Kubernetes
+> documentation](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes)
+> and the [design
+> document](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1698-generic-ephemeral-volumes)
+> for more information about generic ephemeral inline volumes.
 
 
 ### Example of inline CSI pod spec
```
msau42, line 45: Do we need an example of generic pod spec?

book/src/ephemeral-local-volumes.md
```diff
@@ -2,15 +2,44 @@
...
+
+Status | Min K8s Version | Max K8s Version
+--|--|--
+Alpha | 1.19 | -
 
 ## Overview
-Traditionally, volumes that are backed by CSI drivers can only be used with a `PersistentVolume` and `PersistentVolumeClaim` object combination. This feature supports ephemeral storage use cases and allows CSI volumes to be specified directly in the pod specification.  At runtime, nested inline volumes follow the ephemeral lifecycle of their associated pods where the driver handles all phases of volume operations as pods are created and destroyed.
+Traditionally, volumes that are backed by CSI drivers can only be used with a `PersistentVolume` and `PersistentVolumeClaim` object combination. The "CSI Ephemeral Inline volumes" feature allows CSI volumes to be specified directly in the pod specification for ephemeral use cases.  At runtime, nested inline volumes follow the ephemeral lifecycle of their associated pods where the driver handles all phases of volume operations as pods are created and destroyed.
+
+CSI drivers have to be modified to support "*CSI* Ephemeral Inline
```
msau42, line 21: What if we make this a new section, "Which feature should my driver support?", and go into details on some use cases and maybe examples.

book/src/ephemeral-local-volumes.md
pohly: That would also change the URL of the page. While I agree that the page name isn't ideal, I'm leaning towards keeping it to avoid breaking links. It only shows up in URLs, so it's not that visible.

book/src/ephemeral-local-volumes.md
pohly: I didn't add it because I wanted to avoid further duplication of content (the Kubernetes user documentation has such an example), but I suppose for the sake of consistency on this page we have to add it. May also help understand the difference, so yes, let's add it.

book/src/ephemeral-local-volumes.md
pohly: I've added that. Please take another look.

PR #398, opened 2020-12-07
storage capacity tracking

pohly:
This is intentionally a very brief overview because the relevant
information is already available in the Kubernetes and
external-provisioner documentation. Here we only need to make CSI
driver developers aware that the feature and that documentation exist.

### DIFF
--- a/book/src/storage-capacity-tracking.md
+++ b/book/src/storage-capacity-tracking.md
@@ -1,0 +1,45 @@
+# Storage Capacity Tracking
+
+## Status
+
+Status | Min K8s Version | Max K8s Version
+--|--|--
+Alpha | 1.19 | -
+
+## Overview
+
+Storage capacity tracking allows the Kubernetes scheduler to make more
+informed choices about where to start pods which depend on unbound
+volumes with late binding (aka "wait for first consumer"). Without
+storage capacity tracking, a node is chosen without knowing whether
+those volumes can be made available for the node. Volume creation is
+attempted and if that fails, the pod has to be rescheduled,
+potentially landing on the same node again. With storage capacity
+tracking, the scheduler filters our nodes which do not have enough
+capacity.
+
+> For design information, see the [enhancement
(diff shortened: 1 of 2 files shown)


book/src/storage-capacity-tracking.md
```diff
@@ -0,0 +1,42 @@
...
+  capacity is exhausted
+- use external-provisioner >= 1.6.0 because older releases did not
+  properly support rescheduling after a `ResourceExhausted` error
+
+To support storage capacity tracking, a CSI driver deployment must:
+- use external-provisioner >= 2.0.0
+- enable producing of storage capacity objects as explained in the
+  [external-provisioner
+  documentation](https://github.com/kubernetes-csi/external-provisioner/tree/release-2.0#capacity-support)
+- enable usage of that information by setting the
```
msau42, line 37: Also enable the alpha feature gate in Kubernetes 1.19+

book/src/storage-capacity-tracking.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Usage
+
+To support rescheduling of a pod, a CSI driver deployment must:
+- return the `ResourceExhausted` gRPC status code in `CreateVolume` if
+  capacity is exhausted
+- use external-provisioner >= 1.6.0 because older releases did not
+  properly support rescheduling after a `ResourceExhausted` error
+
+To support storage capacity tracking, a CSI driver deployment must:
```
msau42, line 32: Also the driver needs to report capacity in the csi GetCapacity call.

book/src/storage-capacity-tracking.md
pohly: Added with wording that fits into the bullet list ("must run on a cluster with the API enabled") and referred to the upstream documentation for details on how to ensure that.

book/src/storage-capacity-tracking.md
pohly: Sometimes one has to point out the obvious :joy:  Added.

PR #406, opened 2021-01-13
[docs/csi] use correct link to kubelet plugin registration mechanism

veshij:
Currently all "kubelet plugin registration mechanism" links point to [device plugin registration](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-registration) page which should not be used by CSI drivers (node-driver-registrar doesn't implement device plugin registration interface).

Was not able to find proper doc link on https://kubernetes.io/docs, so using https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/pluginmanager/pluginwatcher/README.md instead.


veshij:
Hi, @msau42, @xing-yang!
What do you think about this change?

msau42:
/lgtm
/approve

Thanks for the fix!

PR #412, opened 2021-02-11
Update guidance on ephemeral csi volumes

mattcary:
/assign @msau42 
/cc @andyzhangx

### DIFF
--- a/book/src/ephemeral-local-volumes.md
+++ b/book/src/ephemeral-local-volumes.md
@@ -32,7 +32,7 @@ However, the two features are targeted at different use cases and thus
 have different APIs and different implementations.
 
 > See the [CSI inline
-> volumes](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190122-csi-inline-volumes.md)
+> volumes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/596-csi-inline-volumes)
 > and [generic ephemeral
 > volumes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1698-generic-ephemeral-volumes)
 > enhancement proposals for design details. The user facing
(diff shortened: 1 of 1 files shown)


book/src/ephemeral-local-volumes.md
```diff
@@ -77,6 +77,11 @@ A CSI driver is suitable for CSI ephemeral inline volumes if:
 - ephemeral inline volumes have to be supported on Kubernetes clusters
   which do not support generic ephemeral volumes
 
+A CSI driver is not suitable for CSI ephemeral inline volumes when:
+- provisioning is not local to the node
```
msau42, line 81: I think we didn't want to necessary limit the feature to only local things as long as it could handle proper pod authentication.

book/src/ephemeral-local-volumes.md
```diff
@@ -77,6 +77,11 @@ A CSI driver is suitable for CSI ephemeral inline volumes if:
...
+- ephemeral volume creation requires administration-level parameters, for
```
msau42, line 82: Another aspect is if the volumeAttributes contains information that would want to be restricted by an administrator (such as volume references when there's no Pod-level authorization)

mattcary: Can you give an example here?

My first thought was to say something like "provisioning is done in the controller and would need to be replicated in nodePublish", but that felt too functional. I think that's roughly the idea, though: if adding csi ephemeral support would require moving/duplicating provisioning code from the controller to the node then there's a bad smell.

But it would be good to be more precise I think.

book/src/ephemeral-local-volumes.md
mattcary: Can you give an example here?

My first thought was to say something like "provisioning is done in the controller and would need to be replicated in nodePublish", but that felt too functional. I think that's roughly the idea, though: if adding csi ephemeral support would require moving/duplicating provisioning code from the controller to the node then there's a bad smell.

But it would be good to be more precise I think.

msau42: There could be some driver that stores per-Pod information in a backend, perhaps like a secret manager.

Agree that if you're driver is already supporting the normal PV method, and you would have to duplicate it here, that's not desirable.

book/src/ephemeral-local-volumes.md
msau42: So with this feature, you can set the volumeAttributes directly in the Pod spec. So that's what I'm referring to here. If any of those attributes are things that an admin would want to restrict, then PV/StorageClass is a better place for it.

In this particular example, the driver has nfs server and share as attributes. Those are things an admin would want to restrict and not allow any Pod to specify anything.

mattcary: updated, ptal

Issue #123, opened 2019-03-04
Define a deprecation policy for csi sidecars


msau42:
It can mirror the Kubernetes policy.

Define policy on:
* sidecar CLI arguments
* sidecar behavior
* compatibility with K8s
* compatibility with CSI spec

This may also tie into defining our sidecar versioning scheme and what warrants a major, minor and patch version bump.

/assign
cc @jsafrane @pohly @saad-ali

jsafrane:
+ RBAC policy changes (e.g. new leader election).

pohly:
"CLI arguments", "RBAC policy" and "CSI spec" can all be considered as "compatibility with existing deployments": if a deployment swaps out the current version of a sidecar with a newer one, will the deployment still work?

"compatibility with K8S" is an extended version of that question: will the deployment still work *on the same clusters* as before?

PR #414, opened 2021-02-28
Add project versioning and support policies

msau42:
Fixes: #123

We've already been following this for 2 years now. It's overdue to formalize it.

One thing I did change is on the topic of multiple minor versions that have the same min, recommended, and max Kubernetes versions. I'm proposing that we simplify the policy and just say we support every minor branch for 1 year. The conditional stance we currently have of dropping support for the minor release if there's a later minor release that is equivalent is confusing to explain and understand.


book/src/project-policies.md
```diff
@@ -0,0 +1,61 @@
+# Versioning, Support, and Kubernetes Compatibility
+
+## Versioning
+
+Each Kubernetes CSI component version is expressed as *x.y.z*, where *x* is the
+major version, *y* is the minor version, and *z* is the patch verseion,
```
pohly, line 6: Typo: s/verseion/version/

book/src/project-policies.md
```diff
@@ -0,0 +1,61 @@
...
+
+Patch version releases only contain bug fixes that do not break any backwards
+compatibility.
+
+Minor version releases may contain new functionality that do not
+break any backwards compatibility (except for alpha features).
+
+Major version releases may contain new functionality or fixes that may break
+backwards compatibility with previous major releases. Changes that require a
+major version increase include: removing or changing API/flags//behavior, new
```
pohly, line 17: Single slash? `API/flags/behavior`

book/src/project-policies.md
```diff
@@ -0,0 +1,61 @@
...
+RBAC requirements that are not opt-in, new Kubernetes minimum version
+requirements.
```
pohly, line 19: It might be worth explaining how this is tested:
> A litmus test for not breaking compatibility is to replace the image of a component in an existing deployment without changing that deployment in any other way.

book/src/project-policies.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Support
+
+The Kubernetes CSI project follows the broader Kubernetes project on support.
+Every minor release branch will be supported with patch releases for 1 year.
```
pohly, line 24: "... will be supported with patch releases *at least* for one year, starting with the first release of that branch."

If there's no newer release branch to replace it, then the latest one will continue to be supported for more than one year.

Given that the revised criteria is now strictly time based, would it make sense to add a column with that release date to the version tables of each sidecar?

For example, when looking at https://kubernetes-csi.github.io/docs/external-snapshotter.html it is not obvious which of the branches are recent enough to still be supported.

book/src/project-policies.md
```diff
@@ -0,0 +1,61 @@
...
+
+Minimum version specifies the lowest Kubernetes version where the component will
+function with the most basic functionality, and no additional features added later.
+Generally, this aligns with the Kubernetes version where that CSI spec version was added.
+
+### Maximum Version
+
+Similarly, the max Kubernetes version generally aligns with when support for
+that CSI spec version was removed or if a particular Kubernetes API or feature
+was deprecated and removed.
```
pohly, line 48: I think we should remove "was deprecated". An API gets deprecated much earlier than the actual removal and therefore sidecars continue to be compatible even after deprecation.

Is "Max version" inclusive or exclusive? From the description, it's exclusive (when a required API is removed, the sidecar no longer works), but this might be worth calling out explicitly to avoid confusion.

pohly (commented):
Looks good overall, some comments on wording.

book/src/project-policies.md
jsafrane: +1 for listing expected end of support.

> If there's no newer release branch to replace it, then the latest one will continue to be supported for more than one year.

This implies we can stop supporting older branches that were released 1 year ago at the point we release a new major version. IMO, we should have some overlap where both the old branch + the new branch are supported, to give driver or Kubernetes vendors time to update everything. With my storage vendor hat on, it should be at least 3 months, preferably even more.

book/src/project-policies.md
msau42: thanks for the suggestion! added

book/src/project-policies.md
msau42: tried to word it in the least confusing way... ptal

book/src/project-policies.md
msau42: it's supposed to be inclusive. I fixed the wording.

book/src/project-policies.md
msau42: Regarding updating the sidecar tables, I agree we should do that and also add the date of first release. But I want to do it in a separate PR, it's going to take a fair amount of time to look everything up.

book/src/project-policies.md
```diff
@@ -1,36 +1,69 @@
...
+component in an existing deployment without changing that deployment in any
+other way.
+
+## Support
+
+The Kubernetes CSI project follows the broader Kubernetes project on support.
+Every minor release branch will be supported with patch releases for at least 1 year,
+starting with the first release of that minor version. In addition, the minor
+release branch will be supported for at least 3 months after the next minor
+version is released, to allow time to integrate with the latest release.
```
pohly, line 31: So if any of these two criteria apply, the branch is still supported? It's one year for release-1.0 since 1.0.0 came out, even if 1.1.0 and release-1.1 get created one month after 1.0.0?

That could lead to a proliferation of branches that all need to be supported even when in practice the newer ones are backward-compatible drop-in replacements for the older ones.

book/src/project-policies.md
```diff
@@ -1,36 +1,69 @@
...
+requirements.
+
+A litmus test for not breaking compatibility is to replace the image of a
+component in an existing deployment without changing that deployment in any
+other way.
+
+## Support
+
+The Kubernetes CSI project follows the broader Kubernetes project on support.
+Every minor release branch will be supported with patch releases for at least 1 year,
```
pohly, line 28: > will be supported with patch releases

It might be worth clarifying further what that means - and then we can check whether we actually live up to that promise.

In practice, "still supported" means that cherry-picks for bug fixes are accepted and new releases tagged if someone cares to create the PRs and asks for a release.

There is no systematic monitoring of bug fixes in master to determine which of those should be backported and to which older releases.

book/src/project-policies.md
msau42: The intent here is to give storage vendors a grace period to adapt and qualify the new release while they still have to support the older releases.

In practice, there's two scenarios I see:
* We continue to regularly make releases every quarter. Then this second clause doesn't apply because the 1 year clause is longer and overlaps. In this scenario, an active component that releases 3 times a year could end up needing to support 3 release branches at any one time.
* A particular component is relatively stable and hasn't had any big changes in a year (it's all been patch releases). The number of release branches that are supported could be as low as 1. When time comes for a new minor release, then in this scenario, we would support 2 release branches for 3 months, after which the older one drops out of support.

book/src/project-policies.md
msau42: added "on as as-needed" basis

book/src/project-policies.md
pohly: I'm okay with that. I just wanted to call out this corner case and be sure that we interpret the description the same way.

pohly:
LGTM, but I'll also let @jsafrane comment before merging.

PR #422, opened 2021-04-01
Add Intelliflash Block Storage driver

goldranjith:
Add Intelliflash Block Storage driver.


xing-yang:
Can you squash your commits into 1?  Can you also change the title of this PR from "Patch 2" to "Add Intelliflash Block Storage driver"?

Issue #419, opened 2021-03-17
Define support policy for major versions


chrishenzie:
Our support policy currently covers minor releases for one year, however it is not clear what is the support policy for older major versions.

Something we could do (for example) is:
> Every major release branch will be supported with minor releases for at least 1 year.

Or potentially:
> Only the most recent major release branch will be supported with minor releases.

This will have implications on how many cherry-picks we need to support once a feature makes it into master. Please feel free to note any other implications this support policy would have.

@msau42 @xing-yang

msau42:
Some of my initial thoughts on this. Major versions are disruptive to consumers, so I do think we need to define some policy that balances the amount of upkeep we need to do maintain multiple major releases and the toil that gets imposed on consumers.

I'm thinking we can extend the "overlap" policy we have for minor releases to major releases as well:
```
In addition, the minor release branch will be supported for at least 3 months after the next minor version is released, to allow time to integrate with the latest release.
```

What this means is when we release a new major version, we also release a new minor version on the previous major.

pohly:
> What this means is when we release a new major version, we also release a new minor version on the previous major.

What would be in that minor version release? This sentence sounds like we promise to do such a release regardless whether we have any relevant content for it or not.

msau42:
My thoughts are there is some set of changes that caused us to bump the major version. The remaining changes, bug fixes and new features that don't depend on those breaking changes are candidates that can be included in the previous major release. If there are no such changes (ie, all changes were reasons to bump the major version), then we don't need to release anything on the previous major.

xing-yang:
> What this means is when we release a new major version, we also release a new minor version on the previous major.

Are you suggesting we always release a new minor version at the time we release a new major version, or if the major version is released less than 3 months after the previous minor version release?  I don't think this should become a "regular" process.  This should be an exception.  We can define conditions that allow such an exception.

> The remaining changes, bug fixes and new features that don't depend on those breaking changes are candidates that can be included in the previous major release

I think this is too broad.  Bug fixes are fine, but we should define what kind of new features are qualified for minor release after a major release.  This issue came up because there is a request to backport metrics related feature.  So we can consider metrics as a type of feature that can qualify for a minor release in this case.

jsafrane:
I think we should not backport features to older minor releases. We're adding them to the current minor release (unless they break the compatibility). Older minor branches should be only for bugfixes.

I agree with @msau42, we can extend the existing statement, I propose:
[...]

PR #423, opened 2021-04-04
Support policy for new minor versions on older major versions

chrishenzie:
See https://github.com/kubernetes-csi/docs/issues/419 for more context.

Any assistance with clarifying the wording is appreciated.

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

### DIFF
--- a/book/src/project-policies.md
+++ b/book/src/project-policies.md
@@ -22,6 +22,14 @@ A litmus test for not breaking compatibility is to replace the image of a
 component in an existing deployment without changing that deployment in any
 other way.
 
+To minimize the number of branches we need to support, we do not have a general
+policy for releasing new minor versions on older majors. We will make exceptions
+for work related to meeting production readiness requirements. Only the previous
+major version will be eligible for these exceptions, so long as the time between
+the previous major version and the current major version is under six months.
+For example, if "X.0.0" and "X+1.0.0" were released under six months apart,
+"X.0.0" would be eligible for new minor releases.
+
 ## Support
 
 The Kubernetes CSI project follows the broader Kubernetes project on support.


chrishenzie:
@msau42 @xing-yang

book/src/project-policies.md
```diff
@@ -22,6 +22,10 @@ A litmus test for not breaking compatibility is to replace the image of a
 component in an existing deployment without changing that deployment in any
 other way.
 
+To minimize the number of branches we need to support, we do not have a general
+policy for releasing new minor versions on older majors. We will make exceptions
+for work related to meeting production readiness requirements.
+
```
xing-yang, line 28: We should add a clause that this older major can only be one version lower than the current major and also the time between the two major releases is less than 3-4 months, etc.

chrishenzie: Done. Does the wording above make sense?

msau42: i'm not sure how to read this. Are we trying to say that if "X.0.0" and "X+1.0.0" were released about a quarter apart, then we will consider some exception?

But if it was "X.2.0" and "X+1.0.0", then we won't consider any exception?

xing-yang: Something like that.  We only look at the initial release times of the major releases for minor release exceptions.  Otherwise we could end up with too many releases.  For example, if X.0.0 was released a year before X+1.0.0, we should move on and not to keep releasing minor releases on X.

chrishenzie: Does the wording in this PR capture this in an understandable way? Or what if we worded it like so?

> Only prior major versions released within four months of the current major version will be eligible for this exception.

This wording does broaden it to more than one major release, but maybe it's more intuitive than the current wording.

msau42: I think the wording as is captures it, but I wanted to confirm the intended scenario. Maybe give an example with numbers to make it more clear.

chrishenzie: Done

chrishenzie: Just checking if this captures everything. Do we also want to add a section for RBAC not requiring new majors?

book/src/project-policies.md
```diff
@@ -22,6 +22,14 @@ A litmus test for not breaking compatibility is to replace the image of a
 component in an existing deployment without changing that deployment in any
 other way.
 
+To minimize the number of branches we need to support, we do not have a general
+policy for releasing new minor versions on older majors. We will make exceptions
+for work related to meeting production readiness requirements. Only the previous
+major version will be eligible for these exceptions, so long as the time between
+the previous major version and the current major version is under four months.
```
msau42, line 29: Thinking about this some more, given that the Kubernetes release cadence is slowling down, I wonder if 4 months is actually enough enough to capture even a single release. Should we bump it to 6 months? @xing-yang

book/src/project-policies.md
msau42: I think RBACs still require some more discussion, so let's consider that separately.

book/src/project-policies.md
xing-yang: Sounds good.

chrishenzie: Done

PR #462, opened 2021-09-08
Update Longhorn compatible CSI version

c3y1huang:
https://github.com/longhorn/longhorn-manager/blob/master/go.mod#L31


c3y1huang:
cc @yasker

yasker:
LGTM.

PR #487, opened 2022-01-04
Allow RBAC changes in minor version releases

chrishenzie:
See https://github.com/kubernetes-csi/docs/issues/419 for context.

```release-note
NONE
```

@msau42 
@xing-yang


chrishenzie:
See https://github.com/kubernetes-csi/docs/issues/419#issuecomment-1020416563

We are not going to proceed with this change.

/close

PR #503, opened 2022-02-23
Add support for overriding hostname

gnufied:
Useful for running `make serve` on a remote machine
using 0.0.0.0 or a different hostname


gnufied:
cc @xing-yang @msau42

PR #558, opened 2023-04-10
NetApp Trident CSI spec support up to 1.8

ameade:


clintonk (approved):

book/src/drivers.md
```diff
@@ -80,7 +80,7 @@ Name | CSI Driver Name | Compatible with CSI Version(s) | Description | Persiste
...
 [MooseFS](https://github.com/moosefs/moosefs-csi) | `com.tuxera.csi.moosefs` | v1.0 | A Container Storage Interface (CSI) Driver for [MooseFS](https://moosefs.com/) clusters. | Persistent | Read/Write Multiple Pods | Yes | 
-[NetApp](https://github.com/NetApp/trident) | `csi.trident.netapp.io` | v1.0, v1.1, v1.2, v1.3, v1.4, v1.5 | A Container Storage Interface (CSI) Driver for NetApp's [Trident](https://docs.netapp.com/us-en/trident/index.html) container storage orchestrator | Persistent | Read/Write Multiple Pods | Yes | Raw Block, Snapshot, Expansion, Cloning, Topology
+[NetApp](https://github.com/NetApp/trident) | `csi.trident.netapp.io` | v1.0, v1.1, v1.2, v1.3, v1.4, v1.5, v1.6, v1.7, v1.8 | A Container Storage Interface (CSI) Driver for NetApp's [Trident](https://docs.netapp.com/us-en/trident/index.html) container storage orchestrator | Persistent | Read/Write Multiple Pods | Yes | Raw Block, Snapshot, Expansion, Cloning, Topology
```
msau42, line 83: Maybe a range would scale better?

I also wonder in general how useful is it to put the supported versions in this table and instead rely on driver documentation? That way drivers don't have to continuously keep updating this page with every new version. @xing-yang @jsafrane wdyt?

xing-yang: Yes, a range would look better.
I'm fine not to have this column.  This column is probably not up to date for most drivers on this page.

ameade: I too would prefer to either not have this column or only list major versions.

PR #595, opened 2024-05-22
Fix typo in csi-proxy.md

styshoo:


niranjandarshann:
@styshoo Thank You for your contribution. Can you see Whether there is more typo in entire page. So that it may be fixed in one go. Else Changes looks good to me.

PR #602, opened 2024-07-19
Import Existing Snapshot Example

micmorris:
This adds a brief page with an example of how to import an existing snapshot from the underlying system into Kubernetes. 

I found the documentation to not be straightforward on this topic, and I think this is a common-enough feature to warrant its own explanation and simple example. Just trying to save someone else time from the fiddling I did!


micmorris:
(I just signed the CLA, if that check needs to be retriggered now)

PR #608, opened 2024-11-06
Update drivers.md

csiblock:
IBM Block storage driver 1.12.0 supports RWX, no more support for A9000


csiblock:
thank you @xing-yang for your assistance!

PR #611, opened 2025-01-02
Recommend actual sidecar installation in sample pod

ebblake:
A kubernetes sidecar is an initContainer with restartPolicy:Always. Since we are already calling the helper repositories sidecars, it helps if the example for the user actually installs them as a true sidecar, rather than as a regular container.


xing-yang:
/hold

I don't think we should make this recommendation unless this has proved to be a better deployment method.

ebblake:
> I don't think we should make this recommendation unless this has proved to be a better deployment method.

What would it take to compare the two styles of deployment to see if there are any benefits or drawbacks between the two?

PR #593, opened 2024-04-26
update PVC annotations secret supported in external-provisioner

hoyho:
As mentioned in https://github.com/kubernetes-csi/external-provisioner/issues/1148, external-provisioner didn't recognize per per volume secret tokens that reference to PVCs. This was fixed is master branch of external-provisioner. CSI document should be updated to advertise this change to users.
However, external-provisioner v4.0.2 hasn't published yet.  This PR should wait until the new version is published.

PR #614, opened 2025-02-17
update PVC annotations secret supported in external-provisioner

nixpanic:
As mentioned in https://github.com/kubernetes-csi/external-provisioner/issues/1148, external-provisioner didn't recognize per per volume secret tokens that reference to PVCs. This was fixed is master branch of external-provisioner. CSI document should be updated to advertise this change to users.

Replaces: #593


nixpanic:
/release-note-none

This is a minor change that was introduced quite a while ago. Also, I don't think the docs have releases at all?

